PR #8, opened 2017-03-15
Initial project structure

dgutride:
Includes:
- webpack for bundling, test and dev configurations
- sample component
- sample unit test
- gulp file for transpilation and build helpers
- starter demo site structure


config/webpack.demo.js
```diff
@@ -0,0 +1,182 @@
+const helpers = require('./helpers');
+const webpack = require('webpack');
+const path = require('path');
+const sass = require('./sass');
```
AdamJ: Is this repo going to use the SASS version of PatternFly, or stick with the default LESS version from V3?

gulpfile.js
```diff
@@ -0,0 +1,105 @@
...
+      return file.base; // because of Angular 2's encapsulation, it's natural to save the css where the less file was
+    }));
+});
+
+//Less compilation and minifiction - adopted from sass compilation, needs work
+gulp.task('transpile-less', function () {
+  return transpileLESS(appSrc + '/app/**/*.less');
+});
+
+// Put the SASS files back to normal
```
AdamJ: Confused here (and related to above question) is this using both preprocessors?

dgutride: It is only going to be LESS - there is another issue to manage CSS (which this does not cover).  @mindreeper2420

config/webpack.demo.js
dgutride: This is an unused variable - see comment below and does not indicate any technology choice.

dlabrecq (approved):
LGTM

PR #22, opened 2017-06-13
Added Filter

dlabrecq:
Added filter component and updated to latest ngx-bootstrap to take advantage of Angular 4.

Note: ng2-bootstrap was renamed ngx-bootstrap

screen shot 2017-06-13 at 1 57 18 pm


dtaylor113 (approved):

src/app/filters/examples/filter-example.component.ts
```diff
@@ -0,0 +1,319 @@
...
+  matchesFilters(item: any, filters: Filter[]): boolean {
+    let matches = true;
+    filters.forEach((filter) => {
+      if (!this.matchesFilter(item, filter)) {
+        matches = false;
+        return false;
+      }
+    });
+    return matches;
+  }
```
jeff-phillips-18: Not sure I understand why 'matches' exists. Why set it to false just to return false?

src/app/filters/examples/filter-example.component.ts
```diff
@@ -0,0 +1,319 @@
...
+      this.filterConfig.fields[index].queries = [
+        ...this.imageQueriesFixed,
+        this.separator,
+        ...this.imageQueries.filter((item: any) => {
+          if (item.value) {
+            return (item.value.toLowerCase().indexOf(val.toLowerCase()) > -1);
+          } else {
+            return true;
+          }
+        })
```
jeff-phillips-18, line 305: return !item.value || (item.value.toLowerCase().indexOf(val.toLowerCase()) > -1);

src/app/filters/examples/filter-example.component.ts
```diff
@@ -0,0 +1,319 @@
...
+      ];
+    } else if (this.filterConfig.fields[index].id === 'weekDay') {
+      this.filterConfig.fields[index].queries = [
+        ...this.weekDayQueries.filter((item: any) => {
+          if (item.value) {
+            return (item.value.toLowerCase().indexOf(val.toLowerCase()) > -1);
+          } else {
+            return true;
+          }
```
jeff-phillips-18, line 314: return !item.value || (item.value.toLowerCase().indexOf(val.toLowerCase()) > -1);

src/app/filters/filter-fields.component.less
```diff
@@ -0,0 +1,47 @@
...
+    height: 20px;
+    margin-right: 5px;
+  }
+}
+
+.input-group {
+  .input-group-btn {
+    .dropdown-menu>.selected>a {
+      background-color: @color-pf-blue !important; // #0088ce
+      border-color: #0076b7 !important;
```
jeff-phillips-18: Any better way to get around this than using !important?
Why #0076b7 and not a color from the palette?

src/app/filters/examples/filter-example.component.ts
dlabrecq: That was taken verbatim from the Angular PF example. Will can clean that up.

src/app/filters/filter-fields.component.ts
```diff
@@ -0,0 +1,121 @@
...
+  ngDoCheck(): void {
+    // Do a deep compare on config
+    if (!_.isEqual(this.config, this.prevConfig)) {
+      this.setupConfig();
+    }
+  }
+
+  setupConfig(): void {
+    this.prevConfig = _.cloneDeep(this.config);
+
```
jeff-phillips-18: What happens when this.config is undefined?

src/app/filters/filter-fields.component.less
dlabrecq: Again, this was taken verbatim from the Angular PF. I'm assuming there was a reason for doing that, but don't know what the issue was? Didn't want to change code without understanding it's history.

The color looks close to @color-pf-blue-400, so I'll use that instead.

src/app/filters/filter-fields.component.ts
dlabrecq: No errors actually occur here. However, there are config properties that are undefined during rendering. 

That said, I've added the '?' operator in the HTML templates to allow the component to be rendered without the config. Doesn't look pretty, but no errors are output.

jeff-phillips-18 (approved):

PR #28, opened 2017-06-20
Fix demo site

dgutride:
Add examples to demo dist
Fix refresh when running
Move to non-html5 mode for demo so refresh works for devs when working

Example here:  https://rawgit.com/dgutride/patternfly-ng/copy-examples-dist/dist-demo/index.html#/


dlabrecq (approved):
LGTM

PR #48, opened 2017-07-05
fix: Semantic release - initial checkin for support

dgutride:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "patternfly-ng",
-  "version": "0.0.5",
+  "version": "0.0.0-semantic-release",
```
dlabrecq, line 3: Can this be 0.0.6? We're already on 0.0.5

.travis.yml
```diff
@@ -12,13 +19,20 @@ env:
 before_install:
   - rvm install 2.3.1
   - if [[ `npm -v` != 3* ]]; then npm i -g npm@3; fi
-  - npm install -g bower grunt-cli
+  - npm install -g bower
```
dlabrecq, line 22: Do we need to install bower?

.travis.yml
```diff
@@ -1,6 +1,13 @@
...
+cache:
+  directories:
+    - node_modules
+
+notifications:
+  email: false
+
 node_js:
-  - '4'
+  - '6'
```
dlabrecq, line 10: We had been using node 4 because of build issues. We can try it, but might need to fall back here?

package.json
dgutride: For semantic release - it needs to be set by the build system and library so we pick a string identifier instead of a number.

.travis.yml
dgutride: Unsure - we can take a look at that as well.   Can we do that after this get in?

.travis.yml
dgutride: Let's give it a try - this was the recommended version from the semantic release cli.

Issue #31, opened 2017-06-22
Unit tests needed


dlabrecq:
Unit tests are needed for notification and wizard components.

dlabrecq:
Closing due to inactivity.

PR #85, opened 2017-08-02
Wizard component

dlabrecq:
Adding the wizard component, based on the Angular Patternfly implementation with some refactoring, bug fixes, etc. For example, patternfly-ng favors emitting events (for functionality such as next, previous, close, etc.) instead of relying on user provided callbacks.

This PR includes:

- Components for wizard, wizard-step, wizard-substep, and wizard-review.
- Classes for wizard-config, wizard-step-config, and wizard-event.
- A wizard-base class for common wizard and wizard-step functionality.
- A wizard-step interface for step and substep types.
- A modal example
- A modal example with paused navigation (a Cloud Forms requirement)
- An embed in page example showing a slightly different button layout

Note: Angular Patternfly does not provide unit tests for wizard. Need to figure out how to test ng-templates properly, but not for this go around -- added to issue https://github.com/patternfly/patternfly-ng/issues/31.

Live Example:
https://rawgit.com/dlabrecq/patternfly-ng/wizard-dist/dist-demo/#/wizard

Modal Example:
screen shot 2017-08-04 at 11 49 12 am

Embed in Page Example:
screen shot 2017-08-04 at 11 50 27 am


dlabrecq:
FYI @sjcox-rh @catrobson

package.json
```diff
@@ -67,8 +67,8 @@
     "@angular/router": "^4.0.1",
     "core-js": "2.4.1",
     "moment": "2.17.1",
-    "ngx-bootstrap": "1.7.1",
-    "patternfly": "^3.24.0",
+    "ngx-bootstrap": "1.8.0",
+    "patternfly": "^3.5.0",
```
AdamJ: should this be version `^3.25.0`?

dlabrecq: Yes, thanks for finding that! Bumped to v3.26.1

dgutride (approved):

AdamJ (approved):

catrobson:
@dlabrecq can we add labels to the steps along the top?  "First Step, Second Step, etc".  It looks like there is space for them but nothing is showing.
image

@mindreeper2420 it looks like the final step of the wizard is using the fontawesome fa-check-circle rather than the patternfly pficon-ok as specified by the pattern.  Is this inherited and needs to be fixed upstream?

I also noticed that in the implementation the final screen shows the cancel and back buttons as disabled, but the pattern shows the cancel button disappearing and the back button enabled.  I think how it is implemented is actually more accurate to normal usage.  Do you all agree?  If so I'll create an issue against PatternFly core.

dlabrecq:
@catrobson I've fixed the labels.

The check circle is actually using the icon below. This is the same icon used in the Patternfly test pages and for Angular Patternfly. It's actually defined by the app developer, not the wizard itself.

&lt;div class="wizard-pf-success-icon">
  &lt;span class="glyphicon glyphicon-ok-circle">&lt;/span>
&lt;/div>

Regarding the cancel button behavior, the example also mimics the Patternfly test pages and Angular Patternfly. However, it's up to the app developer to define that behavior. Thus, I can change the example to do whatever you prefer?

dlabrecq:
@catrobson @mindreeper2420 What would you like to do about the ok icon and cancel button?

catrobson:
@dlabrecq I'm good to merge as is.  OK icon should be fixed as we get new versions of PF-icons in the CSS, and Cancel button works as expected for an example IMO.

AdamJ (approved):

PR #110, opened 2017-08-15
feat(sparkline): add sparkline chart

dgutride:


src/app/chart/sparkline/sparkline.component.ts
```diff
@@ -0,0 +1,224 @@
...
+   * Chart data for the chart
+   */
+  @Input() chartData: any;
+
+  /**
+   * Configuration object containing details about how to render the chart
+   */
+  @Input() config: SparklineConfig;
+  private prevChartData: any;
+  public sparklineChartId: any;
```
dlabrecq: Do we want this to be public?

src/app/chart/sparkline/sparkline.component.ts
```diff
@@ -0,0 +1,224 @@
...
+      this.updateAll();
+    }
+
+    if (!isEqual(this.config, this.prevConfig)) {
+      this.setupConfig();
+      this.generateChart(this.sparklineChartId, true);
+    }
+  }
+
+  protected setupConfig(): void {
```
dlabrecq: Looking at the typedoc for wizard and protected methods still need comments.

src/app/chart/sparkline/sparkline.component.ts
```diff
@@ -0,0 +1,224 @@
...
+     * By default only line and the tick marks are shown, no labels. This is a sparkline and should be used
+     * only to show a brief idea of trending. This can be overridden by setting the config.axis options per C3
+     */
+    if (this.config.axis) {
+      this.config.axis.x.show = this.config.showXAxis === true;
+      this.config.axis.y.show = this.config.showYAxis === true;
+    }
+  }
+
+  protected updateAll(): void {
```
dlabrecq: Needs comment. Or, make it private?

src/app/chart/sparkline/sparkline.component.ts
```diff
@@ -0,0 +1,224 @@
...
+    // Need to deep watch changes in chart data
+    this.prevChartData = cloneDeep(this.chartData);
+    // Convert the given data to C3 chart format
+    this.config.data = merge(this.config.data, this.getSparklineData(this.chartData));
+  }
+
+  /*
+   * Convert the config data to C3 Data
+   */
+  protected getSparklineData(chartData: any): any {
```
dlabrecq: Needs comment. Or, make it private?

src/app/chart/sparkline/sparkline.component.ts
```diff
@@ -0,0 +1,224 @@
...
+      sparklineData.columns = [
+        chartData.xData,
+        chartData.yData
+      ];
+    }
+
+    return sparklineData;
+  };
+
+  protected getTooltipTableHTML(tipRows: any): string {
```
dlabrecq: Needs comment. Or, make it private?

dlabrecq (approved):
Looks good. Just need to add some comments to the protected methods.

dgutride:
Here's the preview for design review:  https://rawgit.com/dgutride/patternfly-ng/sparkline-dist/dist-demo/index.html#/sparkline

src/app/chart/chart.base.ts
```diff
@@ -0,0 +1,45 @@
...
+   */
+  @Input() public config: ChartConfig;
+
+  /**
+   * Event emitted with the chart reference after load is complete
+   * @type {EventEmitter}
+   */
+  @Output() chartLoaded: EventEmitter<any> = new EventEmitter();
+
+  public prevConfig: any;
```
dlabrecq: This is showing up in the typedoc as public. The component appears to use prevChartData, so it looks like this can be private.

src/app/chart/chart.base.ts
```diff
@@ -0,0 +1,45 @@
...
+  // store the chart object
+  private chart: any;
+
+  constructor() {
+  }
+
+  generateChart (chartId: string, reload?: boolean): void {
```
dlabrecq: This should probably be protected or private. It's showing up in typedoc without comments.

src/app/chart/sparkline/sparkline.component.ts
dlabrecq: This is showing up in the typedoc as public -- looks like it should be private?

src/app/chart/sparkline/sparkline.component.ts
```diff
@@ -0,0 +1,224 @@
...
+  /**
+   *  Check if the component config has changed
+   */
+  ngDoCheck(): void {
+    // Do a deep compare on config
+    if (!isEqual(this.chartData, this.prevChartData)) {
+      this.updateAll();
+    }
+
+    if (!isEqual(this.config, this.prevConfig)) {
```
dlabrecq, line 69: prevConfig appears to be only updated upon ngOnInit when  super.generateChart is called.

Do we want to call super.ngOnInit here? Otherwise, maybe we should compare prevChartData to see if that has changed -- prevConfig probably won't change after ngOnInit, right?

package.json
```diff
@@ -66,6 +66,7 @@
     "@angular/platform-browser-dynamic": "^4.0.1",
     "@angular/platform-server": "^4.0.1",
     "@angular/router": "^4.0.1",
+    "c3": "^0.4.15",
```
AllenBW, line 69: Drive by observation, we might want to lock down this version number, just in case something magically changes and breaks everything 🤔  (also many of the other deps are locked down, but some are not? 🤔 🤔 )

dlabrecq:
@catrobson and/or @mindreeper2420 Would you please take a quick look at the visuals?

catrobson:
@dgutride is there a way to turn off the area filling and just have the line itself as well?

dlabrecq:
We will have a separate line chart component for that. For example:

http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfLineChart

catrobson (approved):
Reviewed from design standpoint - approved.  Will put note about design contribution needed into PatternFly notes.

Issue #114, opened 2017-08-18
Wizard's config to hideFooter


corinnekrych:
As we have hideHeader, in some wizard we need to hideFooter (the change could be dynamic) allow full size windows for the main content.
In https://github.com/patternfly/patternfly-ng/blob/master/src/app/wizard/examples/wizard-basic-example.component.ts#L107 replace by:
```
     this.wizardConfig = {
       title: 'Wizard Title',
       sidebarStyleClass: 'example-wizard-sidebar',
       stepStyleClass: 'example-wizard-step',
      hideFooter: true
     } as WizardConfig;
 ```
to enable hideFooter option.

dlabrecq:
This would allow the wizard to be used as a modal, but it would break the PatterfnFly wizard pattern.

Since #115 has been closed, I'll close this issue as well.

PR #115, opened 2017-08-18
feat(wizard): be able to hide footer in wizard's config

corinnekrych:
Fixes https://github.com/patternfly/patternfly-ng/issues/114

### DIFF
--- a/src/app/wizard/wizard-config.ts
+++ b/src/app/wizard/wizard-config.ts
@@ -45,6 +45,11 @@ export class WizardConfig {
   hideHeader?: boolean;
 
   /**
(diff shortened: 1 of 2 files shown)


dlabrecq:
I spoke with @catrobson about OSIO's usage of the wizard. I don't believe we want to hide the footer (i.e., the back, next, cancel buttons). 

Does the OSIO design not have the footer?

catrobson:
Agreed that we don't want to hide this, but use it in our wizard for the cancel, next, finish button area of the current wizard.

dlabrecq:
Spoke with @corinnekrych offline and we agreed to close this.

PR #118, opened 2017-08-22
fix(filterdropdown): Added nbsp to filter tooltip.

michaelkleinhenz:
Added a nbsp in the filter dropdown tooltip so it does not wrap.

### DIFF
--- a/src/app/filter/filter-fields.component.html
+++ b/src/app/filter/filter-fields.component.html
@@ -2,7 +2,7 @@
   <div class="input-group form-group">
     <div class="input-group-btn" dropdown>
       <button type="button" class="btn btn-default filter-fields dropdown-toggle" dropdownToggle
-              tooltip="Filter by" placement="{{config?.tooltipPlacement}}">
+              tooltip="Filter&nbsp;by" placement="{{config?.tooltipPlacement}}">
         {{currentField?.title}}
         <span class="caret"></span>
       </button>


src/app/filter/filter-fields.component.html
```diff
@@ -2,7 +2,7 @@
   <div class="input-group form-group">
     <div class="input-group-btn" dropdown>
       <button type="button" class="btn btn-default filter-fields dropdown-toggle" dropdownToggle
-              tooltip="Filter by" placement="{{config?.tooltipPlacement}}">
+              tooltip="Filter&nbsp;by" placement="{{config?.tooltipPlacement}}">
```
AdamJ, line 5: why not apply `white-space: nowrap` to the button instead?

https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

dlabrecq: If you don't have enough room for the tooltip, you could add your own CSS. That is, using Adam's suggestion or by setting width.

See the example given here:
https://github.com/valor-software/ngx-bootstrap/issues/808

dlabrecq:
Actually, I see the wrapping in our example, which has plenty of room to display the tooltip. We should fix it here.

dlabrecq:
The following change would resolve the issue. 

https://github.com/patternfly/patternfly-ng/pull/119

@michaelkleinhenz Do you want to update or close?

PR #120, opened 2017-08-26
Tree List component

dlabrecq:
Adding the tree list component. This component makes use of angular-tree-component to provide a tree hierarchy along with drag and drop support, lazy loading of child nodes, etc.

https://github.com/500tech/angular-tree-component

Note that this is a somewhat different component than the original version created for ngx-widgets. For example, it's now one component instead of two, which simplifies things for the end user.

The ngx-widgets component is also based on v2.7.0 of angular-tree-component and Angular 2. This new tree list component uses the latest v4.1.0 of angular-tree-component, which is based on Angular 4. The latest angular-tree-component also uses a different HTML template layout, so some porting/customizing was necessary.

This new tree list component closely mimics the existing patternfly-ng list component, which uses the latest pf-list layout from PatternfFly core. Both list and tree list components are able to share many of the same features such as checkbox selection, single row (click) selection, multiple row selection, etc. These features are not available with the ngx-widgets component.

Since the list and tree list components have so much in common, I've refactored the existing list component a bit and moved common functionality to base classes for the two components to share. Note that the existing list component class names have not changed.

Example:
https://rawgit.com/dlabrecq/patternfly-ng/tree-list-dist/dist-demo/#/tree-list

screen shot 2017-08-26 at 3 50 40 am


dlabrecq:
FYI, @catrobson

AdamJ:
LGTM from CSS/HTML design end.

AdamJ (approved):

src/app/list/basic-list/examples/list-compound-example.component.html
```diff
@@ -98,10 +98,10 @@ <h4 class="events-label">Actions: </h4>
         <iframe class="demoframe" src="docs/classes/listcomponent.html"></iframe>
       </tab>
       <tab heading="html">
-        <include-content src="src/app/list/examples/list-compound-example.component.html"></include-content>
+        <include-content src="src/app/list/list/examples/list-compound-example.component.html"></include-content>
```
dgutride: Should this be basic-list, too?

src/app/list/basic-list/examples/list-heading-example.component.html
```diff
@@ -141,10 +139,10 @@ <h4 class="events-label">Actions: </h4>
         <iframe class="demoframe" src="docs/classes/listcomponent.html"></iframe>
       </tab>
       <tab heading="html">
-        <include-content src="src/app/list/examples/list-heading-example.component.html"></include-content>
+        <include-content src="src/app/list/list/examples/list-heading-example.component.html"></include-content>
```
dgutride: basic-list...

dgutride (changes requested):

src/app/list/basic-list/examples/list-compound-example.component.html
dlabrecq: Yes. Thanks!

dlabrecq:
Reviewed the demo with Adam and fixed the HTML includes Dana pointed out. Should be good to go now.

dgutride (approved):

PR #123, opened 2017-08-27
Added a lazy filter example

dlabrecq:
This additional filter example shows how to add filters lazily. See the new "Lazy Loading" tab.

FYI, this is to help the Planner team.

Example:
https://rawgit.com/dlabrecq/patternfly-ng/filter-dist/dist-demo/#/filters

### DIFF
--- a/src/app/filter/examples/filter-lazy-example.component.html
+++ b/src/app/filter/examples/filter-lazy-example.component.html
@@ -3,7 +3,21 @@
     <div class="col-sm-12">
       <div class="form-group">
         <pfng-filter [config]="filterConfig"
-                     (onChange)="filterChanged($event)"></pfng-filter>
+                     (onChange)="filterChanged($event)"
+                     (onFieldSelect)="filterFieldSelected()"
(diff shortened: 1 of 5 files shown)


src/app/filter/examples/filter-lazy-example.component.html
```diff
@@ -0,0 +1,79 @@
...
+  </div>
+  <div class="row padding-bottom-15 padding-top-15">
+    <div class="col-xs-12">
+      <h4>Valid Items:</h4>
+      <hr/>
+    </div>
+  </div>
+  <div class="row">
+    <div class="col-md-12">
+      <div *ngFor="let item of items" class="col-md-12">
```
AdamJ, line 32: not sure if a 12 grid column inside of a 12 grid column is necessary?

AdamJ (approved):
There is a mix of `col-xs-*`, `col-sm-*` and `col-md-*`, which might make more sense to be consistent when they are only 12 columns.

dgutride (approved):

PR #125, opened 2017-08-29
fix(filter): Do not emit filter add even when filter value is empty

### LINKED ISSUE #124: Filter should not be added when field value is empty
In Angular PatternFly:
https://github.com/patternfly/angular-patternfly/blob/a07d2c78263e6dc41f342eaf041d47096b9c6dcf/src/filters/simple-filter/filter-fields-component.js#L92-L93

While in PatternFly-ng:
https://github.com/patternfly/patternfly-ng/blob/5535abd79692aade228b4d20ca4ea5636e7daa82/src/app/filter/filter-fields.component.ts#L147-L151

astefanutti:
fixes #124

### DIFF
--- a/src/app/filter/filter-fields.component.ts
+++ b/src/app/filter/filter-fields.component.ts
@@ -144,7 +144,7 @@ export class FilterFieldsComponent implements DoCheck, OnInit {
   // Private
 
   private fieldInputKeyPress($event: KeyboardEvent): void {
-    if ($event.which === 13) {
+    if ($event.which === 13 && this._currentValue && this._currentValue.length > 0) {
       this.onAdd.emit({
         field: this._currentField,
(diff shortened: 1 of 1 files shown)


dlabrecq (commented):
We will need to test this still works with type ahead functionality.

dlabrecq (approved):
This change was just introduced to Angular Patternfly on 8/23.

See: https://github.com/patternfly/angular-patternfly/pull/593

Although patternfly-ng does not generate a similar null.legnth error, it does allow an empty filter.

astefanutti:
Is there anything else that should be done on my side?

dlabrecq:
Just needed two people to review. Thanks for the fix!

astefanutti:
Thanks for the new release 👍

Issue #131, opened 2017-09-06
Inconsistent Caret in toolbar dropdown


SMahil:
In toolbar dropdown caret icon is incosistent

Steps to Reproduce:
Please refer the video link: https://drive.google.com/a/redhat.com/file/d/0BzAb-YONh7fsbWdiQ1BJT0lvSlk/view?usp=sharing

dlabrecq:
This is not an issue with the toolbar, but the filter fields component. 

There are different types of filters; text, select, and typeahead. The typeahead is a new feature, which was recently contributed. The 'select' type uses a button to display its underlying menu, the 'typeahead' uses div tags. 

It's the 'filter-select' style used with the button that sets font-style to italic and its color to #8b8d8f. The 'typeahead-input-container' style could/should also set the font and color to match.

PR #132, opened 2017-09-06
Inconsistent filter caret styling

dlabrecq:
The filter caret styling isn't consistent for the types 'select' and 'typeahead'. They should both italic fonts.

Fixes https://github.com/patternfly/patternfly-ng/issues/131

screen shot 2017-09-06 at 1 59 34 pm

Example:
https://rawgit.com/dlabrecq/patternfly-ng/filter-131-dist/dist-demo/#/toolbar

### DIFF
--- a/src/app/filter/filter-fields.component.less
+++ b/src/app/filter/filter-fields.component.less
@@ -14,9 +14,11 @@
       position: relative;
       padding-right: 0;
       .caret {
+        color: @color-pf-black-500;
+        font-style: italic;
         position: absolute;
         top: 10px;
-        right: 5px;
+        right: 12px;
         z-index: 2;
       }
     }


AdamJ:
@dlabrecq the caret is still different between the two, based off of your link
screen shot 2017-09-11 at 11 45 20 am

dlabrecq:
Yes, but all I can really do is make it italic as they are two different elements.

dlabrecq:
@mindreeper2420 are you speaking about the drop shadow for the dropdown menu? If so, the "week Day" example is showing an input element, while the "Birth Month" example is using a Bootstrap dropdown.

Input elements don't typically don't have the drop shadow like menus.

AdamJ:
@dlabrecq not the drop shadow, but that the caret in the 'Filter by Birth Month' is italic, but the caret for 'Birth Month' is not.

dlabrecq:
Ok, I see what you're describing now. The filter field is not italic, but the input place holder text is italic.

We're not actually changing the caret for the filter field. Only the place holder text for the filter query (i.e., the dropdown and input), which has it's own caret. 

That is, we're just making the new type ahead feature match the existing italic place holder text.

AdamJ (approved):
Sounds good - thanks for the explanation!

dgutride (approved):

PR #137, opened 2017-09-21
Contribute vertical navigation component

dgutride:
Demo here: 
http://rawgit.com/dgutride/patternfly-ng/vertical-navigation-dist/dist-demo/#/navigation


index.ts
```diff
@@ -87,6 +87,11 @@ export { ToolbarComponent } from './src/app/toolbar/toolbar.component';
 export { ToolbarModule } from './src/app/toolbar/toolbar.module';
```
dlabrecq, line 87: 1. Unlike the Angular Patternfly example, clicking on the various tabs does not seem to navigate anywhere? If we're not supporting the basic example, I would like to see an example of router navigation?
2. Clicking on the "Exit Demo" tab doesn't work. (For the Angular Patternfly, clicking on the "Exit Demo" tab functions.) I didn't understand that users must click on a button in the header, at first -- thought that was a tooltip.
3. Text is cut off when the hamburger icon is toggled
4. No icon for the "Exit Demo" tab when the hamburger icon is toggled -- the tab is blank
5. The "PatternFly-ng" title should appear centered vertically

screen shot 2017-09-27 at 1 01 46 pm

src/app/navigation/examples/simple-vertical-navigation-example.component.ts
```diff
@@ -0,0 +1,286 @@
...
+  ChangeDetectorRef,
+  Component,
+  OnInit
+} from '@angular/core';
+import { NavigationStart, Router } from '@angular/router';
+import { NavigationItemConfig } from '../navigation-item-config';
+
+@Component({
+  selector: 'simple-vertical-navigation-example',
+  styles: [`
```
dlabrecq: Should we move the styles to their own file for consistency? The other examples in patternfly-ng use files, but I don't feel strongly about it.

src/app/navigation/examples/simple-vertical-navigation-example.component.ts
```diff
@@ -0,0 +1,286 @@
...
+      right: 0;
+      background-color: #f5f5f5;
+      padding-top: 15px;
+    }
+
+    .hide-vertical-nav {
+      margin-top: 15px;
+      margin-left: 30px;
+    }
+    /*.sample-form .form-horizontal .form-group {*/
```
dlabrecq: We should remove the commented-out code

src/app/navigation/examples/simple-vertical-navigation-example.component.ts
```diff
@@ -0,0 +1,286 @@
...
+    this.showExample = !this.showExample;
+    this.chRef.detectChanges();
+  }
+
+  onItemClicked($event): void {
+    this.actionText += 'Item Clicked ' + '\n';
+  }
+
+  onNavigation($event): void {
+    this.actionText += 'Navigation event fired '  + '\n';
```
dlabrecq: This is a little confusing because I don't see the action text when I click on a tab. I only see that a navigation event was fired after exiting the demo. Perhaps it would help if you could append the name of the tab that was clicked?

src/app/navigation/examples/vertical-navigation-example.component.html
```diff
@@ -0,0 +1,16 @@
...
+    </div>
+  </div>
+  <tabset>
+    <tab heading="Basic" (select)="tabSelected($event)">
+      <simple-vertical-navigation-example *ngIf="activeTab === 'Basic' || activeTab === ''"></simple-vertical-navigation-example>
+    </tab>
+    <tab heading="Router" (select)="tabSelected($event)">
+      <simple-vertical-navigation-example *ngIf="activeTab === 'Router'"></simple-vertical-navigation-example>
+    </tab>
+  </tabset>
```
dlabrecq: I don't see tabs in the example. Are we supporting both basic and router navigation? I thought you said something about only supporting router navigation?

screen shot 2017-09-27 at 1 49 18 pm

src/app/navigation/examples/simple-vertical-navigation-example.component.ts
```diff
@@ -0,0 +1,286 @@
+import {
```
dlabrecq: Can this be renamed basic-vertical-navigation-example.component.ts for consistency? We've been using that "basic-" as a sudo naming convention with other examples (e.g., basic card, basic list, basic wizard, etc.). Really nothing simple about these components. ;)

src/app/navigation/examples/vertical-navigation-example.component.ts
```diff
@@ -0,0 +1,31 @@
+import {
```
dlabrecq, line 1: Is this the router navigation example? Will we support both basic and router examples? If so, can we add 'router' to the example name for clarity?

src/app/navigation/navigation-item-config.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Title for the navigation item
+   */
+  title: string;
+
+  /**
+   * The icon class to use for icons displayed to the left of text
+   */
+  iconClass?: string;
```
dlabrecq: Can we rename this as iconStyleClass for consistency with other components?

src/app/navigation/navigation-item-config.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Link to navigate to
+   */
+  href?: string;
```
dlabrecq: Thinking this should be named 'url' to be consistent with other component attribute names.

src/app/navigation/navigation-item-config.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Navigation children (used for secondary and tertiary navigation)
+   */
+  children?: NavigationItemConfig[];
+
+  /**
+   * Indicate if the item should be active on load
+   */
+  active?: boolean;
```
dlabrecq: Thinking this should be named activeOnLoad considering there is also an isActive boolean.

src/app/navigation/navigation-item-config.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Track the active state of the navigation item
+   */
+  isActive?: boolean;
```
dlabrecq: Don't care for the isActive and isHover names. Although I've seen the 'is' prefixed for some attribute names, we've reserved this convention for methods (i.e., something that tests state where a boolean is returned).

I'm thinking trackActiveState and trackHoverState may be more appropriate here?

src/app/navigation/navigation-item-config.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Indicates if the child tertiary menu is opened
+   */
+  tertiaryCollapsed?: boolean;
+
+  /**
+   * Indicates if this is a mobile item
+   */
+  isMobileItem?: boolean;
```
dlabrecq: Thinking we should drop the 'is' prefix and use mobileItem

src/app/navigation/navigation-item-config.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Internal variable used for hovering timeout
+   */
+  navHoverTimeout?: any;
+
+  /**
+   * Internal variable used for unhovering timeout
+   */
+  navUnHoverTimeout?: any;
```
dlabrecq: I've never seen an 'unhover' state, so 'blur' or 'onmouseleave' may be more appropriate? And no need for the 'nav' prefix as everything here is for the nav component. I'm thinking hoverTimeout and blurTimeout?

src/app/navigation/vertical-navigation.component.ts
```diff
@@ -0,0 +1,665 @@
...
+
+  /**
+   * Hide top banner optionally
+   */
+  @Input() hideTopBanner: boolean;
+
+  /**
+   * This event is fired any time the user has initiated navigation
+   */
+  @Output('onNavigatedEvent') navigatedEvent = new EventEmitter();
```
dlabrecq: I'd like to see this named onNavigationEvent. I've been trying to ovoid names with plurals, past tense, etc.

src/app/navigation/vertical-navigation.component.ts
```diff
@@ -0,0 +1,665 @@
...
+    this.activeSecondary = false;
+    this.showMobileNav = false;
+    this.showMobileSecondary = false;
+    this.showMobileTertiary = false;
+    this.hoverSecondaryNav = false;
+    this.hoverTertiaryNav = false;
+    this.collapsedSecondaryNav = false;
+    this.collapsedTertiaryNav = false;
+    this.navCollapsed = false;
+    this.forceHidden = false;
```
dlabrecq: Can these defaults be declared on lines 83-95? That is, unless the component is being reinitialized, it would be cleaner.

src/app/navigation/vertical-navigation.component.ts
```diff
@@ -0,0 +1,665 @@
...
+
+    this.routeChangeListener = this.router.events.subscribe((val) => {
+      if(val  instanceof NavigationEnd) {
```
dlabrecq: The TS linter picks this up as a spacing issue.

src/app/navigation/vertical-navigation.component.ts
```diff
@@ -0,0 +1,665 @@
...
+    this.navCollapsed = false;
+
+    //Set the body class to the correct state
+    this.removeClass('collapsed-nav');
+
+    this.explicitCollapse = false;
+
+    // Dispatch a resize event when showing the expanding then menu to
+    // allow content to adjust to the menu sizing
+    window.dispatchEvent(new Event('resize'));
```
dlabrecq: Would like to create a window service, since we're supporting mobile -- like we have for codebases in OSIO. We also reference window in the chart and action components, so it could be used there as well.

"Referencing global browser objects like document or window directly from within your code is possible, but not encouraged and considered bad practice"

"Especially Angular isn’t only designed to run within your browser, but also on mobiles, the server or web workers where objects like window may not be available. Therefore the suggested approach is to wrap such objects and inject them through the dependency injection mechanism. This way it is possible to change the concrete runtime instance of a given object based on the environment the Angular application is running."

src/app/navigation/vertical-navigation.component.ts
```diff
@@ -0,0 +1,665 @@
...
+          }
+        });
+      }
+    });
+  }
+
+  /**
+   * Sets the active items in the vertical navigation
+   */
+  public setActiveItems(): void {
```
dlabrecq: Typically a 'set' method takes an argument. Perhaps renaiming this as initActiveItems would work better?

src/app/navigation/vertical-navigation.component.ts
```diff
@@ -0,0 +1,665 @@
...
+        });
+      }
+    });
+  }
+
+  /**
+   * Public resize event called when the window size changes
+   * @param event
+   */
+  public onResize(event: any): void {
```
dlabrecq, line 506: Does this or any of the following methods need to be public? (I typically try to make everything private, then see what AOT complains about.) I'm asking because the TS linter is complaining about the public methods appearing after the private ones.

I also built the code and added it to my Angular CI app. There are a few errors regarding private methods. For example:

ERROR in ng:///Users/dlabrecq/Work/patternfly/test/patternfly-ng-notification/node_modules/patternfly-ng/src/app/navigation/vertical-navigation.component.d.ts.VerticalNavigationComponent.html (2,722): Property 'activeSecondary' is private and only accessible within class 'VerticalNavigationComponent'.

src/app/navigation/vertical-navigation.component.ts
```diff
@@ -0,0 +1,665 @@
...
+        }
+      }
+    }
+  }
+
+  /**
+   * Hides menus on unhover
+   * @param item
+   */
+  public handlePrimaryUnHover(item: NavigationItemConfig): void {
```
dlabrecq: There's that non-existent 'unhover' state, mentioned above, in the method name. Would like to rename this as handlePrimaryBlur or handlePrimaryOnMouseLeave?

dlabrecq (changes requested):

dlabrecq (changes requested):
Just one issue left. When the mouse leaves the nav, it remains open unlike the Angular Patternfly basic vertical nav example.

I believe the menu should only remain open if the user clicks on the pin?

screen shot 2017-10-05 at 11 07 52 am

dlabrecq (approved):

PR #139, opened 2017-09-29
Add LESS linter

### LINKED ISSUE #84: Add Less Linter
Add a Less Linter to the build for .less files.

Suggested linter: https://www.npmjs.com/package/gulp-lesshint

recrsn:
Adds [gulp-lesshint](https://www.npmjs.com/package/gulp-lesshint) to the gulp build process.

I have not specified any style configuration, so the default configuration generates a handful of warnings now.

Closes #84

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -8,6 +8,7 @@ var gulp = require('gulp'),
   fs = require("fs"),
   htmlMinifier = require('html-minifier'),
   lessCompiler = require('gulp-less'),
+  lessHint = require('gulp-lesshint'),
   ngc = require('gulp-ngc'),
   path = require('path'),
   postcss = require('postcss'),
@@ -52,6 +53,12 @@ function updateWatchDist() {
     .src([libraryDist + '/**'].concat(globalExcludes))
     .pipe(changed(watchDist))
     .pipe(gulp.dest(watchDist));
+}
+
+function lintLESS(src) {
+  return gulp.src(src)
+    .pipe(lessHint())
+    .pipe(lessHint.reporter());
 }
 
 function transpileLESS(src) {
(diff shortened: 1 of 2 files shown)


dlabrecq (changes requested):
This generates a ton of warnings.

screen shot 2017-10-05 at 11 26 08 am

I think we really need a style configuration file before we can close #84?

OSIO uses stylelint and stylelint-webpack-plugin. I believe we closely follow their style config, so perhaps their stylelintrc file would help here?

recrsn:
@dlabrecq Should we use [gulp-stylelint](https://github.com/olegskl/gulp-stylelint) instead ? In that case we can just reuse the style configuration of OSIO ?

dlabrecq:
Eventually, we will also duplicate OSIO's TS and HTML linters, so it may be best to add them all to webpack?

I was thinking we would duplicate what OSIO has done in webpack.common.config? Then, copy ..stylelintignore and .stylelintrc to this repo.

      const StyleLintPlugin = require('stylelint-webpack-plugin');
      ...
      /*
       * StyleLintPlugin
       */
      new StyleLintPlugin({
        configFile: '.stylelintrc',
        syntax: 'less',
        context: 'src',
        files: '**/*.less',
        failOnError: true,
        quiet: false,
      })

recrsn:
@dgutride By the look of our gulpfile, it seems we are processing less into css before in enters webpack. Or, am I mistaken ?

dlabrecq:
You're correct, we're processing the Less before webpack in order to in-line the CSS. This is necessary to support AOT compilation. We could probably move that to webpack, but not something we must tackle now.

If you can use gulp-stylelint with OSIO's style config, let's do that instead?

Issue #32, opened 2017-06-22
Port pagination component from Angular Patternfly


dlabrecq:
Support and/or example needed for pagination.

See: http://www.patternfly.org/pattern-library/navigation/pagination/
See: http://www.patternfly.org/angular-patternfly/#/api/patternfly.pagination.component:pfPagination

sahil143:
@dlabrecq I'm working on this.

PR #151, opened 2017-10-05
feat(pagination): added pagination Component

sahil143:
This component is based on `pf-pagination` from angular-patternfly repo.

Issue: #32 

https://rawgit.com/sahil143/patternfly-ng/pagination-dist/dist-demo/#/pagination


sahil143:
@dlabrecq @dgutride Please review this component.

index.ts
```diff
@@ -63,6 +63,11 @@ export { ToastNotificationComponent } from './src/app/notification/toast-notific
 export { ToastNotificationListComponent } from './src/app/notification/toast-notification-list.component';
```
dlabrecq, line 63: AOT compilation does not work. 

screen shot 2017-10-05 at 1 48 24 pm

As I mentioned via email, you need to create an Angular CLI app, install your own build of patternfly-ng, and compile with 'ng build --prod'. Then, adjust the component as necessary to complete a good AOT build.

src/app/pagination/examples/pagination-example.component.ts
```diff
@@ -0,0 +1,45 @@
+import {
+  Component,
+  OnInit,
+  ViewEncapsulation
+} from '@angular/core';
+
+import { PaginationConfig } from '../pagination-config';
+import { PaginationComponent } from '../pagination.component';
```
dlabrecq: This is unused and should be removed.

src/app/pagination/examples/pagination-example.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  data() {
+    for (let i = 1; i < 126; i++) {
+      this.items.push({ id: i, status: 'Ok', value: Math.floor(Math.random() * (1000 - 1 + 1)) + 1 });
+    }
+  }
+
+  handlePageSize(event) {
+    this.actionsText ="Page Size: " + event.pageSize + " Selected" + "\n" + this.actionsText;
```
dlabrecq: There a a couple issues with the linter here.

1. No space after equals symbol
2. Change " char to '

src/app/pagination/examples/pagination-example.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  handlePageNumber(event) {
+    this.actionsText ="Page Number: " + event.pageNumber + " Selected" + "\n" + this.actionsText;
```
dlabrecq: Same linter issue as line 38.

src/app/pagination/examples/pagination-example.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+}
```
dlabrecq: No return char

src/app/pagination/pagination-config.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  /**
+   * The current page number
+   */
+  pageNumber?: number;
+
+  /**
+   * The total number of items in the data set.
+   */
+  numTotalItems?: number;
```
dlabrecq: Thinking this should be renamed as totalItems.

src/app/pagination/pagination.component.html
```diff
@@ -0,0 +1,31 @@
+<div class="content-view-pf-pagination list-view-pf-pagination clearfix">
+  <div class="form-group">
+    <div class="padding-right-10">
+      <div class="btn-group dropdown" dropdown>
+        <button class="btn btn-default dropdown-toggle" dropdownToggle>
+        {{config.pageSize}}
+        <span class="caret"></span>  
+      </button>
+        <ul class="dropdown-menu" *dropdownMenu>
+          <li *ngFor="let increment of config.pageSizeIncrements" [ngClass]="{'selected': increment === config.pageSize}" class="display-length-increment">
```
dlabrecq: Good practice to use a ? char for config properties. For example: config?.pageSize, config?.pageSizeIncrements -- this ensures no errors when a value is not provided.

I should not see any errors when the config object is not provided; however, I see 'NaN' for the page size.

screen shot 2017-10-05 at 2 36 58 pm

src/app/pagination/pagination.component.html
```diff
@@ -0,0 +1,31 @@
...
+    </ul>
+    <input type="text" #pageNoInput [ngModel]="config.pageNumber" (keyup.enter)="onPageNumberChange(pageNoInput.value)" class="pagination-pf-page"
+      name="page" />
+    <span>of <span class="pagination-pf-pages"> {{ lastPageNumber }} </span></span>
+    <ul class="pagination pagination-pf-forward">
+      <li [ngClass]="{'disabled': config.pageNumber === lastPageNumber}"><a title="Next Page" (click)="gotoNextPage()" class="goto-next-page"><span class="i fa fa-angle-right"></span></a></li>
+      <li [ngClass]="{'disabled': config.pageNumber === lastPageNumber}"><a title="Last Page" (click)="gotoLastPage()" class="goto-last-page"><span class="i fa fa-angle-double-right"></span></a></li>
+    </ul>
+  </div>
+</div>
```
dlabrecq: Missing return char

src/app/pagination/pagination.component.html
```diff
@@ -0,0 +1,31 @@
...
+        <ul class="dropdown-menu" *dropdownMenu>
+          <li *ngFor="let increment of config.pageSizeIncrements" [ngClass]="{'selected': increment === config.pageSize}" class="display-length-increment">
+            <a role="menuitem" (click)="onPageSizeChange(increment)">{{increment}}</a></li>
+        </ul>
+      </div>
+    </div>
+    <span class="per-page-label">per page</span>
+  </div>
+  <div class="form-group">
+    <span><span class="pagination-pf-item-current">{{getStartIndex()}} - {{getEndIndex()}}</span> of <span class="pagination-pf-items-total">{{config.numTotalItems}}</span></span>
```
dlabrecq: For debugging purposes, I would prefer to use a getCurrentPage() method, rather than to make the calculation here.

src/app/pagination/pagination.component.html
```diff
@@ -0,0 +1,31 @@
...
+    <ul class="pagination pagination-pf-back">
+      <li [ngClass]="{'disabled': config.pageNumber === 1}"><a title="First Page" (click)="gotoFirstPage()" class="goto-first-page"><span class="i fa fa-angle-double-left"></span></a></li>
+      <li [ngClass]="{'disabled': config.pageNumber === 1}"><a title="Previous Page" (click)="gotoPreviousPage()" class="goto-prev-page"><span class="i fa fa-angle-left"></span></a></li>
+    </ul>
+    <input type="text" #pageNoInput [ngModel]="config.pageNumber" (keyup.enter)="onPageNumberChange(pageNoInput.value)" class="pagination-pf-page"
```
dlabrecq: I believe we may want to use Angular's onchange event here instead of keyup.enter? That way, an event is only generated if the value changed.

And instead of using onPageNumberChange(pageNoInput.value), use a two way model binding. For example, [(ngModel)]="config.pageNumber", then use (change)="onPageNumberChange($event)" where MouseEvent is passed as the $event object. The component can easily access config.pageNumber internally.

src/app/pagination/pagination.component.html
```diff
@@ -0,0 +1,31 @@
...
+      name="page" />
+    <span>of <span class="pagination-pf-pages"> {{ lastPageNumber }} </span></span>
```
dlabrecq: Are extra spaces necessary here? If so, consider using '\&nbsp;' instead.

src/app/pagination/pagination.component.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+describe('Pagination component - ', () => {
+  let comp: PaginationComponent;
+  let fixture: ComponentFixture<PaginationComponent>;
+  let config: PaginationConfig;
+
+  beforeEach(() => {
+    config = {
+      pageSize: 10,
+      pageNumber: 1,
+      numTotalItems: 125  //explicitly assigned
```
dlabrecq: linter spacing issue with the comment

src/app/pagination/pagination.component.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      });
+  }));
+
+  it('should go to next page', () => {
+    let button = fixture.debugElement.query(By.css('.goto-next-page'));
+    button.triggerEventHandler('click', null);
+    expect(comp.config.pageNumber).toEqual(2);
+  });
+
+  it('should go to previous page', ()=>{
```
dlabrecq: linter spacing issue

src/app/pagination/pagination.component.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    expect(comp.config.pageNumber).toEqual(4);
+  });
+
+  it('should jump to last page', () => {
+    let button = fixture.debugElement.query(By.css('.goto-last-page'));
+    button.triggerEventHandler('click', null);
+    expect(comp.config.pageNumber).toEqual(comp.lastPageNumber);
+  });
+
+  it('should jump to first page', ()=>{
```
dlabrecq: linter spacing issue

src/app/pagination/pagination.component.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    comp.config.pageNumber = 6;
+    fixture.detectChanges();
+    let button = fixture.debugElement.query(By.css('.goto-first-page'));
+    button.triggerEventHandler('click',null);
+    expect(comp.config.pageNumber).toEqual(1);
+  });
+});
```
dlabrecq: Missing return char

src/app/pagination/pagination.component.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      .compileComponents()
+      .then(() => {
+        fixture = TestBed.createComponent(PaginationComponent);
+        comp = fixture.componentInstance;
+        comp.config = config;
+        fixture.detectChanges();
+      });
+  }));
+
+  it('should go to next page', () => {
```
dlabrecq, line 40: We should test the menu functionality as well as the input field. Must ensure users can enter a value directly and invoke the menu properly.

src/app/pagination/pagination.component.ts
```diff
@@ -0,0 +1,205 @@
...
+  @Output('onPageNumberSelect') onPageNumberSelect = new EventEmitter();
+
+
+  private defaultConfig = {
+    pageNumber: 1,
+    pageSizeIncrements: [5, 10, 20, 40, 80, 100],
+    pageSize: 5
+  } as PaginationConfig
+  private prevConfig: PaginationConfig;
+  lastPageNumber: number;
```
dlabrecq: This should be marked private

src/app/pagination/pagination.component.ts
```diff
@@ -0,0 +1,205 @@
...
+    this.onPageSizeSelect.emit({
+      pageSize: newPageSize
+    });
+  }
+
+  /**
+   * Page number is changed using input field
+   * @param value new page number
+   */
+  private onPageNumberChange(value: any): void {
```
dlabrecq: This should not be type 'any'. Input values are type 'string'.

src/app/pagination/pagination.component.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * The Pagination config contaning component properties
+   */
+  @Input() config: PaginationConfig;
+
+  /**
+   * The Event is emitted when Page Size is changed
+   */
+  @Output('onPageSizeSelect') onPageSizeSelect = new EventEmitter();
```
dlabrecq: I'm thinking this should be renamed as onPageSizeChange.

src/app/pagination/pagination.component.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * The Event is emitted when Page Number is Changed
+   */
+  @Output('onPageNumberSelect') onPageNumberSelect = new EventEmitter();
```
dlabrecq: Thinking this should be renamed as onPageNumberChange.

If we had events for buttons like next, prev, etc., then I would probably have used onNextPageSelect, onPreviousPageSelect; for example.

src/app/pagination/examples/pagination-example.component.html
```diff
@@ -0,0 +1,59 @@
+<div class="padding-15">
+  <div class="row padding-bottom-15">
+    <div class="col-xs-12">
+      <h4>Pagination Component Example</h4>
+      <hr/>
+    </div>
+  </div>
+  <div class="row">
+    <div class="col-xs-12">
+      <div *ngFor="let item of items | slice:(paginationConfig.pageNumber - 1)*paginationConfig.pageSize:paginationConfig.numTotalItems | slice:0:paginationConfig.pageSize"
```
dlabrecq: The linter is picking up some spacing issues here. Wrapping the line would solve this and make it more readable; for example:

&lt;div class="col-md-12"
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;*ngFor="let item of items |
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;slice:((paginationConfig.pageNumber - 1) * paginationConfig.pageSize):paginationConfig.numTotalItems |
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;slice:0:paginationConfig.pageSize"&gt;

src/app/pagination/pagination.component.html
```diff
@@ -0,0 +1,31 @@
...
+        </ul>
+      </div>
+    </div>
+    <span class="per-page-label">per page</span>
+  </div>
+  <div class="form-group">
+    <span><span class="pagination-pf-item-current">{{getStartIndex()}} - {{getEndIndex()}}</span> of <span class="pagination-pf-items-total">{{config.numTotalItems}}</span></span>
+    <ul class="pagination pagination-pf-back">
+      <li [ngClass]="{'disabled': config.pageNumber === 1}"><a title="First Page" (click)="gotoFirstPage()" class="goto-first-page"><span class="i fa fa-angle-double-left"></span></a></li>
+      <li [ngClass]="{'disabled': config.pageNumber === 1}"><a title="Previous Page" (click)="gotoPreviousPage()" class="goto-prev-page"><span class="i fa fa-angle-left"></span></a></li>
```
dlabrecq: For completeness, I'd like to see gotoFirstPage($event), gotoPreviousPage($event), gotoNextPage($event), gotoLastPage($event), and onPageSizeChange($event, increment) where MouseEvent is provided as the $event parameter. Even if the MouseEvent object is not ultimately used, it's clear that a mouse click invoked these functions.

src/app/pagination/pagination.component.html
```diff
@@ -0,0 +1,31 @@
...
+        {{config.pageSize}}
+        <span class="caret"></span>  
+      </button>
+        <ul class="dropdown-menu" *dropdownMenu>
+          <li *ngFor="let increment of config.pageSizeIncrements" [ngClass]="{'selected': increment === config.pageSize}" class="display-length-increment">
+            <a role="menuitem" (click)="onPageSizeChange(increment)">{{increment}}</a></li>
+        </ul>
+      </div>
+    </div>
+    <span class="per-page-label">per page</span>
```
dlabrecq: For accessibility, we should add for="pageSizeMenu" to the 'per page' label span element. In this case, I'm thinking #pageSizeMenu would be added to the button element.

src/demo/app.component.html
```diff
@@ -22,6 +22,9 @@
         <li>
           <a routerLink="/emptystate" routerLinkActive="active">Empty State</a>
         </li>
+        <li>
+          <a routerLink="/pagination" routerLinkActive="active">Pagination</a>
```
dlabrecq: Please move the 'Pagination' link so it appears alphabetically in the example list

src/app/pagination/examples/pagination-example.component.html
```diff
@@ -0,0 +1,59 @@
+<div class="padding-15">
```
dlabrecq, line 1: A couple people have asked for a pagination example that demos a list component. I'd like to rename this file as pagination-basic-example.component.html and add a tab, so we can show a simple list example (e.g., pagination-list-example.component.html).

If you're not up for creating that, it can wait for another PR?

dlabrecq (changes requested):
Nice work. Just a few minor issues to address.

sahil143:
@dlabrecq 
* Is there a way to check linter error? Because when  I build It exits successfully...no linter error

> For completeness, I'd like to see gotoFirstPage($event), gotoPreviousPage($event), gotoNextPage($event), gotoLastPage($event), and onPageSizeChange($event, increment) where MouseEvent is provided as the $event parameter. Even if the MouseEvent object is not ultimately used, it's clear that a mouse click invoked these functions.


* passing `$event` as parameter throws AOT error. If we also pass a parameter inside class method then it works fine but then we would have unused parameters passed back and forth between class methods.
* Marking `lastPageNumber` `private` throws AOT errors.

> I believe we may want to use Angular's onchange event here instead of keyup.enter? That way, an event is only generated if the value changed.
And instead of using onPageNumberChange(pageNoInput.value), use a two way model binding. For example, [(ngModel)]="config.pageNumber", then use (onchange)="onPageNumberChange($event)" where MouseEvent is passed as the $event object. The component can easily access config.pageNumber internally.

* `[(ngModel)]="config.pageNumber"`, with `(onchange)="onPageNumberChange($event)"` changes the page even when we press backspace(some undefined values are being shown) and when I add another number(invalid) in `input`

* `onPageSizeChange` in place of `onPageSizeSelect` is not possible because we have method with same name `onPageSizeChange()`.

src/app/pagination/examples/pagination-example.component.html
sahil143: I'm working on this. I have a work in progress branch for this 
https://github.com/sahil143/patternfly-ng/tree/pagination-list/src/app/pagination/examples

dlabrecq:
Regarding the linter error, the build won't fail, but you can clearly see the error in an IDE. I use Webstorm, for example.

For the AOT error, make _lastPageNumber private and add an accessor. For example:

&nbsp;&nbsp;get lastPageNumber(): number

Regarding the backspace, you can test MouseEvent for the key pressed and perform an action based on the key pressed (i.e., only if the enter key was pressed). The only problem with keyup.enter is that your function is called regardless if the value actually changed.

It would look something like this:

<input type=text (onchange)="eventHandler($event)">

eventHandler($event: MouseEvent) {
&nbsp;&nbsp;if ($event.keyCode === 13) {
&nbsp;&nbsp;&nbsp;&nbsp;console.log($event, $event.keyCode, $event.keyIdentifier);
&nbsp;&nbsp;}
}

Regarding the onPageSizeSelect event; yes, it is possible to rename that. If there is already  onPageSizeChange method, simply rename that method to something else, like onPageSizeClick. 

I prefer to use onPageSizeChange as the event name because the user has not selected anything here, like a menu.

sahil143:
@dlabrecq 
* I have added the List example.
* Its still not working for me 
   - when we use one-way data binding i.e. `[ngModel]="config.pageNumber"` and `on-change=onPageNumberUpdate($event, input.value)` I have to pass it the input value because ngModel does not reflect the changes back to `config.PageNumber`.....  but works fine because it only changes the page number and updates `config.pageNumber` when enter is pressed. (this is the one I have used in code)
  - When we use two-way data binding i.e.  `[(ngModel)]=config.PageNumber` and `on-change=onPageNumberUpdate($event, input.value)` it instantly reflects the changes back due to which page number change on backspace and on `config.pageNumber` more than the total number of pages ....  this is happening because first, it performs the two-way binding( change the page number and updates `config.pageNumber`) then it performs `on-change`(execute onpageNumberUpdate() after enter is pressed and updates `config.pageNumber` again).
  - Travis is failing because one of the unit test depend on input.

src/app/pagination/pagination.component.html
```diff
@@ -0,0 +1,32 @@
...
+    </div>
+    <span for="pageSizeMenu" class="per-page-label">per page</span>
+  </div>
+  <div class="form-group">
+    <span><span class="pagination-pf-item-current">{{getCurrentPage()}}</span> of <span class="pagination-pf-items-total">{{config.totalItems}}</span></span>
+    <ul class="pagination pagination-pf-back">
+      <li [ngClass]="{'disabled': config.pageNumber === 1}"><a title="First Page" (click)="gotoFirstPage()" class="goto-first-page"><span class="i fa fa-angle-double-left"></span></a></li>
+      <li [ngClass]="{'disabled': config.pageNumber === 1}"><a title="Previous Page" (click)="gotoPreviousPage()" class="goto-prev-page"><span class="i fa fa-angle-left"></span></a></li>
+    </ul>
+    <input type="text" #pInput [ngModel]="config.pageNumber" on-change="onPageNumberUpdate($event, pInput.value)" class="pagination-pf-page"
```
dlabrecq: You're mixing Angular 1 syntax with Angular 4. Try using (onchange)="onPageNumberUpdate..." instead. Also use a two way binding ([ngModel])="config.pageNumber" so you don't need to pass the value into onPageNumberUpdate.

src/app/pagination/pagination.component.ts
```diff
@@ -0,0 +1,222 @@
...
+    return Math.ceil(this.config.totalItems / this.config.pageSize);
+  }
+
+  /**
+   * Check if current Page is Last Page
+   */
+  private isLastPage(): boolean {
+    return this.config.pageNumber === this.lastPageNumber;
+  }
+}
```
dlabrecq: Missing return at the end of file.

src/demo/app.component.html
```diff
@@ -97,4 +100,4 @@
       <router-outlet></router-outlet>
     </div>
   </div>
-</main>
+</main>
```
dlabrecq: Missing return char

src/app/pagination/examples/pagination-basic-example.component.html
```diff
@@ -0,0 +1,51 @@
...
+  <div>
+    <tabset>
+      <tab heading="api">
+        <iframe class="demoframe" src="docs/classes/paginationcomponent.html"></iframe>
+      </tab>
+      <tab heading="html">
+        <include-content src="src/app/pagination/examples/pagination-example.component.html"></include-content>
+      </tab>
+      <tab heading="typescript">
+        <include-content src="src/app/pagination/examples/pagination-example.component.ts"></include-content>
```
dlabrecq: I'm not seeing the HTML or Typescript doc. Looks like the file names are incorrect?

src/app/pagination/pagination-config.ts
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * Page size increments for the 'per page' dropdown
+   */
+  pageSizeIncrements?: Array<number>;
+
+  /**
+   * The initial page size to use
+   */
+  pageSize?: number;
+}
```
dlabrecq: Missing return char

src/app/pagination/pagination.component.ts
```diff
@@ -0,0 +1,222 @@
...
+    if (this.config !== undefined) {
+      defaults(this.config, this.defaultConfig);
+    } else {
+      this.config = cloneDeep(this.defaultConfig);
+    }
+    this.prevConfig = cloneDeep(this.config);
+  }
+
+  /**
+   * get _lastPageNumber
```
dlabrecq: These comments show up in the API doc, so "get _lastPageNumber" isn't appropriate. You could say something more like "Returns last page number"?

src/app/pagination/pagination.component.ts
```diff
@@ -0,0 +1,222 @@
...
+   */
+  get lastPageNumber(): number {
+    return this._lastPageNumber;
+  }
+
+  /**
+   * set _lastPageNumber
+   */
+  set lastPageNumber(value: number) {
```
dlabrecq: Same as above. Needs a more descriptive comment.

src/app/pagination/pagination.component.ts
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Jump to Last Page
+   */
+  public gotoLastPage(): void {
+    if (this.config.pageNumber < this.lastPageNumber) {
+      this.updatePageNumber(this.lastPageNumber);
+    }
+  }
+
+  public getCurrentPage() {
```
dlabrecq: Missing API doc

src/app/pagination/examples/pagination-list-example.component.html
```diff
@@ -0,0 +1,93 @@
...
+       <pfng-list [actionTemplate] = "actionTemplate"
+                  [config]="listConfig" 
+                  [expandTemplate]="expandTemplate" 
+                  [items]="items" 
+                  [itemTemplate]="itemTemplate"
+                  (onActionSelect)="handleAction($event)" 
+                  (onClick)="handleClick($event)" 
+                  (onDblClick)="handleDblClick($event)" 
+                  (onSelectionChange)="handleSelectionChange($event)">
+                  <ng-template #itemTemplate
```
dlabrecq: Thank you for adding the list example. 

Let's simplify this example a bit. For example, we're not using this:

[expandTemplate]="expandTemplate" 
(onDblClick)="handleDblClick($event)"

src/app/pagination/examples/pagination-list-example.component.ts
```diff
@@ -0,0 +1,329 @@
...
+  templateUrl: './pagination-list-example.component.html'
+})
+export class PaginationListExampleComponent implements OnInit {
+  actionsText: string = '';
+  allItems: any[];
+  emptyStateConfig: EmptyStateConfig;
+  items: any[];
+  itemsAvailable: boolean = true;
+  listConfig: ListConfig;
+  selectType: string = 'checkbox';
```
dlabrecq: Remove selectType -- we don't have any selections in this demo.

src/app/pagination/examples/pagination-list-example.component.ts
```diff
@@ -0,0 +1,329 @@
...
+      title: 'No Items Available',
+      info: 'This is the Empty State component. The goal of a empty state pattern is to provide a good first ' +
+      'impression that helps users to achieve their goals. It should be used when a list is empty because no ' +
+      'objects exists and you want to guide the user to perform specific actions.',
+      helpLink: {
+        hypertext: 'List example',
+        text: 'For more information please see the',
+        url: '#/list'
+      }
+    } as EmptyStateConfig;
```
dlabrecq: We're not showing an empty state, so emptyStateConfig can be removed.

src/app/pagination/examples/pagination-list-example.component.ts
```diff
@@ -0,0 +1,329 @@
...
+
+@Component({
+  encapsulation: ViewEncapsulation.None,
+  selector: 'pagination-list-example',
+  templateUrl: './pagination-list-example.component.html'
+})
+export class PaginationListExampleComponent implements OnInit {
+  actionsText: string = '';
+  allItems: any[];
+  emptyStateConfig: EmptyStateConfig;
```
dlabrecq: Remove emptyStateConfig

src/app/pagination/examples/pagination-list-example.component.ts
```diff
@@ -0,0 +1,329 @@
...
+      helpLink: {
+        hypertext: 'List example',
+        text: 'For more information please see the',
+        url: '#/list'
+      }
+    } as EmptyStateConfig;
+
+    this.listConfig = {
+      dblClick: false,
+      emptyStateConfig: this.emptyStateConfig,
```
dlabrecq: Remove emptyStateConfig

src/app/pagination/examples/pagination-list-example.component.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  /**
+   * Get the ActionConfig properties for each row
+   *
+   * @param item The current row item
+   * @param actionButtonTemplate {TemplateRef} Custom button template
+   * @param startButtonTemplate {TemplateRef} Custom button template
+   * @returns {ActionConfig}
+   */
+  getActionConfig(item: any, actionButtonTemplate: TemplateRef<any>,
```
dlabrecq: If we use the actionTemplate from the compound expansion example, then all the code in the getActionConfig method can be removed.

See my comments above.

src/app/pagination/examples/pagination-list-example.component.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  handleSelectionChange($event: ListEvent): void {
+    this.actionsText = $event.selectedItems.length + ' items selected\r\n' + this.actionsText;
+  }
+
+  handleClick($event: ListEvent): void {
+    this.actionsText = $event.item.name + ' clicked\r\n' + this.actionsText;
+  }
+
+  handleDblClick($event: ListEvent): void {
```
dlabrecq: Unused method, please remove.

src/app/pagination/examples/pagination-list-example.component.html
```diff
@@ -0,0 +1,93 @@
...
+                  <strong>{{item.nodeCount}}</strong> Nodes
+                </div>
+                <div>
+                  <span class="pficon pficon-image"></span>
+                  <strong>{{item.imageCount}}</strong> Images
+                </div>
+              </div>
+            </div>
+          </ng-template>
+          <ng-template #actionTemplate let-item="item" let-index="index">
```
dlabrecq, line 43: Please use the actionTemplate from the compound expansion example. It's far more simplified.

&lt;ng-template #actionTemplate let-item="item" let-index="index">
&nbsp;&nbsp;&nbsp;&nbsp;&lt;pfng-action class="list-pf-actions"
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[config]="actionConfig"
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;(onActionSelect)="handleAction($event, item)">
&nbsp;&nbsp;&nbsp;&nbsp;&lt;/pfng-action>
&lt;/ng-template>

src/app/pagination/examples/pagination-list-example.component.ts
```diff
@@ -0,0 +1,329 @@
...
+    this.actionsText = "Page Number: " + event.pageNumber + " Selected" + "\n" + this.actionsText;
+  }
+
+  // Row selection
+
+  updateItemsAvailable(): void {
+    this.items = (this.itemsAvailable) ? cloneDeep(this.allItems) : [];
+  }
+
+  updateSelectionType(): void {
```
dlabrecq: Unused method, please remove.

src/app/pagination/pagination.component.ts
```diff
@@ -0,0 +1,222 @@
...
+    this.gotoFirstPage();
+    this.onPageSizeChange.emit({
+      pageSize: newPageSize
+    });
+  }
+
+  /**
+   * Page number is changed using input field
+   */
+  public onPageNumberUpdate($event: KeyboardEvent, value: string): void {
```
dlabrecq: Per my previous review, you still need to check the event to ensure the user pressed the enter key.

Please see my earlier comments -- I believe that I provided an example?

src/app/pagination/pagination.component.ts
```diff
@@ -0,0 +1,222 @@
...
+        this.updatePageNumber(1);
+      } else {
+        this.updatePageNumber(newPageNumber);
+      }
+  }
+
+  /**
+   * Jump to First Page
+   */
+  public gotoFirstPage(): void {
```
dlabrecq: Public methods must appear before private methods, per the linter. You can also drop the 'public' prefix as this is already the default for methods and variables.

src/app/pagination/pagination.component.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    // click on menu option with value 20
+    item[2].triggerEventHandler('click',null);
+    expect(comp.config.pageSize).toEqual(20);
+  });
+
+  it('should change page by using input', () => {
+    let input = fixture.debugElement.query(By.css(".pagination-pf-page"));
+    input.nativeNode.value = 6;
+    input.triggerEventHandler('keyup.enter', null);
+    expect(comp.config.pageNumber).toEqual(6);
```
dlabrecq: This test is failing -- please fix

dlabrecq (changes requested):
If the Travis build is failing for unit tests, then the tests fail locally as well.

Try using input.nativeElement instead of input.nativeNode. The tests must pass prior to us being able to merge

src/app/pagination/pagination.component.html
dlabrecq: We could use (keyup.enter)=""; however, that will always generate an event regardless if the value changed -- whenever enter is pressed.

I was also testing the Angular Patternfly example and noticed that they use the blur event. I assume that was done for the case where the user changes the page number, does not press enter, and navigates away from the input field. If pagination isn't changed, the input field would be left with an incorrect value. 

We should ensure that test case is handled as well -- the change an blur events will do that, but keyup.enter probably will not?

In either case, we will still want to use [(ngModel)] instead of passing the input value to onPageNumberChange. I tried this out with your code and it appears to work as expected.

The two way binding will also allow devs to set the current page, via the config.pageNumber, if desired.

dlabrecq:
Entered "a2" in the input field and see "NaN" as the result -- "2a" was handled, tho.

dlabrecq:
@sahil143 Per your request, I've pushed changes to fix various code review issues. 

- ngModel cannot bind directly to config.pageNumber; otherwise, pagination is updated as user types
- Prevent bad characters from being entered ('a2', '2a', etc.) -- current page should be restored
- Removed unused ngOnChanges method
- Fixed blur and keyup events to mimic Angular Patternfly functionality
- Fixed linter (double quote, missing return char, etc.) issues

The changes can be found here:
https://github.com/patternfly/patternfly-ng/pull/151/commits/3adc6082e9ab807e740e03ec0709491b0967dae3

Note that I did not address the unit test that was commented-out, please fix that so we can merge?

FYI, the typed KeyboardEvent object is not including the key code for some reason? However, if we were to use type any, it appears to be provided. Instead, I used a combination of blur and keyup to achieve the desired functionality. The down side of that is pressing enter will always generate an event, even if the value has not changed. We can address that later if it becomes an issue.

sahil143:
@dlabrecq Thanks for the fix.

Yes! typed KeyboardEvent object does not provide keyCode access for (change) event and also for type any it doesn't. 

I have added two more tests one is for blur and other is for keyup.enter. As for Keyup.enter I used a spy to see if the method onPageNumberKeyup is called.

Do I need to Squash commits?

dlabrecq:
For code reviews, it's best not to squash commits after the PR has been created -- it's easier to see the history of what changed. 

Although Github allows us to squash and merged, I'm going to leave it as is.

dlabrecq (approved):
LGTM

PR #155, opened 2017-10-15
Save filter feature

dlabrecq:
Add functionality to support save filter feature from https://github.com/fabric8-ui/fabric8-ux/issues/664

Example: Click on "Save" tab
https://rawgit.com/dlabrecq/patternfly-ng/save-filter-dist/dist-demo/#/filters

Screen shot 1:
screen shot 2017-10-14 at 11 29 06 pm

Screen shot 2:
screen shot 2017-10-14 at 11 29 51 pm

Recorded demo:
https://gfycat.com/gifs/detail/UnluckyUnrulyDuckbillcat

unluckyunrulyduckbillcat-size_restricted

### DIFF
--- a/src/app/filter/examples/filter-save-example.component.html
+++ b/src/app/filter/examples/filter-save-example.component.html
@@ -6,7 +6,7 @@
                      [config]="filterConfig"
                      (onChange)="filterChanged($event)"
                      (onDelete)="deleteSavedFilter($event)"
-                     (onFieldSelect)="filterFieldSelected()"
+                     (onFieldSelect)="filterFieldSelected($event)"
                      (onSave)="saveFilter($event)"
                      (onTypeAhead)="filterQueries($event)"></pfng-filter>
         <div class="padding-top-15">
--- a/src/app/filter/examples/filter-save-example.component.html
+++ b/src/app/filter/examples/filter-save-example.component.html
@@ -2,10 +2,25 @@
   <div class="row">
     <div class="col-sm-12">
       <div class="form-group">
-        <pfng-filter [config]="filterConfig"
+        <pfng-filter #filter
+                     [config]="filterConfig"
                      (onChange)="filterChanged($event)"
-                     (onFieldSelect)="filterFieldSelected($event)"
+                     (onFieldSelect)="filterFieldSelected()"
+                     (onRemove)="removeSavedFilter($event)"
+                     (onSave)="saveFilter($event)"
                      (onTypeAhead)="filterQueries($event)"></pfng-filter>
+        <div class="padding-top-15">
+          This example lazily loads filters in addition to saving a custom filter.
+        </div>
+      </div>
+    </div>
+  </div>
+  <div class="row">
+    <div class="col-md-12">
+      <div class="padding-top-15">
+        <button class="btn btn-default primary-action" type="button"
+                (click)="loadSavedFilters()">Load Saved Filters
+        </button>
       </div>
     </div>
   </div>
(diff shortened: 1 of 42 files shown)


dlabrecq:
FYI, @catrobson

dgutride (dismissed):

catrobson:
@dlabrecq is there any way to do more of a confirmation on the deletion of a filter?  Not sure if that gets too complex for the pattern.  It seems pretty easy to accidentally delete a filter without a "are you sure" step. 

Also, can we add a label "Name your filter" on the field in the save dropdown? 
image

dlabrecq:
Regarding the delete filter, we probably could add something. However, the user can also delete all filters by clicking on the "clear all" link. We don't have a confirmation there, either.

A label is easy. Or, we can do place holder text?

dlabrecq:
@catrobson I added a label to the save filter input. Please let me know if you want the label on top, instead? Using placeholder text would also save some room.

screen shot 2017-10-20 at 10 11 07 am

dlabrecq:
@catrobson I wasn't able to add a confirmation popover (similar to the "save filter") to the remove filter button. The menu items are already wrapped in an anchor, which prevents the Angular Bootstrap popover from displaying there. And the remove filter icon must be located within the anchor in order for the hover state to display properly.

I can display the popover elsewhere, but looks odd not having the arrow attached to the menu option. If there is a different design you would like to consider, a different popup perhaps, let me know?

dlabrecq:
Instead of a confirmation popover, I was able to change the remove filter icon and tooltip when selected. This forces the user to confirm by clicking the new icon. We can use a different icon and tooltip, but please let me know if something like this will work?

If the user selects a different filter to remove, or whenever a filter is applied, the original icon is restored.

Remove filter:
screen shot 2017-10-20 at 1 19 33 pm

Confirm remove filter:
screen shot 2017-10-20 at 1 20 02 pm

catrobson:
Label on top for the filter naming would be better.  The validation check to remove a filter doesn't follow typical standards, but presents some interesting alternatives.  How you have it with a tooltip doesn't feel quite right to me, let me try out one or two things as alternatives and post them back here.

dlabrecq:
Added Brian's slide-in design for the delete filter confirmation. May need some minor tweaks?

Example: Click on "Save" tab
https://rawgit.com/dlabrecq/patternfly-ng/save-filter-dist/dist-demo/#/filters

Trash can:
screen shot 2017-10-28 at 5 55 00 am

Slide-in confirmation:
screen shot 2017-10-28 at 5 54 41 am

Save filter label on top:
screen shot 2017-10-28 at 5 55 29 am

AdamJ:
Clicking on the trash can and then having to move my mouse to the left to click the checkmark feels weird. Would it be possible to swap the checkmark and 'X', so that the user does not have to move their mouse?

dlabrecq:
Yes, I've swapped the x and the check mark. Brian mentioned the same. The demo has been updated.

dgutride (dismissed):

src/app/filter/examples/filter-save-example.component.html
```diff
@@ -0,0 +1,82 @@
...
+    </div>
+  </div>
+  <div class="row padding-bottom-15 padding-top-15">
+    <div class="col-xs-12">
+      <h4>Valid Items:</h4>
+      <hr/>
+    </div>
+  </div>
+  <div class="row">
+    <div class="col-md-12">
```
AdamJ, line 34: Not sure how this will work, with this being `col-md-` and other pieces being `col-xs-`

src/app/filter/filter-fields.component.less
```diff
@@ -50,3 +50,66 @@
...
 }
 /* styleint-enable */
+
+.pfng-filter-delete {
+  font-size: 12px;
+  opacity: 0;
+  padding-top: 4px;
+  .dropdown-item:hover & {
+    opacity: 1;
+    transition-duration: 1s;
```
AdamJ: I'd move this down to `.5s`, as 1 second feels like it is taking too long.

src/app/filter/filter-results.component.less
```diff
@@ -3,3 +3,36 @@
...
+  padding-top: 3px;
+}
+
+.pfng-save-filter {
+  input {
+    display: inline;
+    @media (min-width: 480px) {
+      width: 10em;
+    }
+    @media (min-width: 768px) {
```
AdamJ: use Bootstrap's native media query variable:
```
@media (min-width: @screen-sm-min) { width: 15em; }
```

AdamJ (changes requested):

AdamJ (approved):

Issue #144, opened 2017-09-30
About modal pattern


catrobson:
Can be converted from angular-patternfly:
http://www.patternfly.org/angular-patternfly/#/api/patternfly.modals.component:pfAboutModal

Pattern:
http://www.patternfly.org/pattern-library/communication/about-modal/

dlabrecq:
This available now

PR #176, opened 2017-11-02
feat(modals): added About Modal component

sahil143:
Based on `pf-about-modal` from angular-patternfly repo

Issue: #144 

https://rawgit.com/sahil143/patternfly-ng/modal-dist/dist-demo/#/aboutmodal


sahil143:
@dlabrecq Please Review and

- Where should I place image in the repo?
- For `onCancel` event I have passed MouseEvent object.
- Do we need `config.isOpen` in component because I don't see any use of it? It had use in the Angular-patternfly repo to open and close modal.
- What should I test in unit tests for this component?

dlabrecq:
For images, please use the existing PatternFly images if possible (e.g., node_modules/patternfly/dist/img/logo-alt.svg). 

Update: You could also provide an image in your example config like so: '//www.patternfly.org/assets/img/logo-alt.svg'.

Instead of MouseEvent, you could create a new ModalEvent or just pass true. I don't see any other components passing the underlying mouse events -- probably easier to change, add properties if needed? For example, inline notification simply provides true when the user clicks to remove a notification.

  public notificationRemove(): void {
    this.hidden = true;
    this.hiddenChange.emit(this.hidden);
  }

Considering how the Angular Bootstrap modal is used, I don't believe we need a config.isOpen property. Functionality such as open, close, etc. should already be exposed via the example component. For example, the wizard demo already has these methods:

  closeModal($event: WizardEvent): void {
    this.modalRef.hide();
  }
  openModal(template: TemplateRef&ltlany>): void {
    this.modalRef = this.modalService.show(template, { class: 'modal-lg' });
  }

Regarding unit tests, please start by porting about-modal.spec.js from Angular PatternFly. Feel free to add more unit tests as you see fit.

index.ts
```diff
@@ -53,6 +53,11 @@ export { ListExpandToggleComponent } from './src/app/list/basic-list/list-expand
 export { TreeListComponent } from './src/app/list/tree-list/tree-list.component';
 export { TreeListConfig } from './src/app/list/tree-list/tree-list-config';
 
+// Modals
+export { AboutModalConfig } from './src/app/modals/about-modal-config';
+export { AboutModalComponent } from './src/app/modals/about-modal.component';
+export { ModalsModule } from './src/app/modals/modals.module';
```
dlabrecq: Please move this from modals to modal

patternfly-ng.module.ts
```diff
@@ -28,6 +29,7 @@ import { WizardModule } from './src/app/wizard/wizard.module';
     EmptyStateModule,
     FilterModule,
     ListModule,
+    ModalsModule,
```
dlabrecq: Please rename from ModalsModule to ModalModule

src/app/modals/about-modal-config.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * The source for the corner graphic
+   */
+  imgSrc?: string;
+
+  /**
+   * Flag indicating that the modal should be opened
+   */
+  isOpen?: boolean;
```
dlabrecq: We can remove isOpen per our discussion

src/app/modals/about-modal.component.html
```diff
@@ -0,0 +1,25 @@
+<div class="about-modal-pf">
+  <div class="modal-header">
+    <button type="button" class="close" (click)="close($event)" aria-hidden="true">
+      <span class="pficon pficon-close"></span>
+    </button>
+  </div>
+  <div class="modal-body">
+    <h1 *ngIf="config.title">{{config.title}}</h1>
+    <div *ngIf="config.productInfo && config.productInfo.length > 0" class="product-versions-pf">
```
dlabrecq: The HTML looks nice. A good test would be to try your example without providing a config. Although the modal may look odd, there should be no errors in the browser console.

src/app/modals/about-modal.component.html
```diff
@@ -0,0 +1,25 @@
...
+      <ng-template *ngIf="template" [ngTemplateOutlet]="template">
+      </ng-template>
+    </div>
+    <div *ngIf="config.additionalInfo" class="product-versions-pf">{{config.additionalInfo}}</div>
+    <div *ngIf="config.copyright" class="trademark-pf">{{config.copyright}}</div>
+  </div>
+  <div class="modal-footer">
+    <img *ngIf="config.imgSrc" [src]="config.imgSrc" alt="{{config.imgAlt}}" />
+  </div>
+</div>
```
dlabrecq: Need a return char.

FYI, I added a TS linter and AOT compilation to patternfly-ng. The build should show these issues automatically, but I may need to tweak some settings?

src/app/modals/about-modal.component.ts
```diff
@@ -0,0 +1,91 @@
...
+   */
+  @Input() template: TemplateRef<any>;
+
+  /**
+   * The Event is emitted when modal is closed
+   */
+  @Output('onCancel') onCancel = new EventEmitter();
+
+  private defaultConfig= {
+    isOpen: false
```
dlabrecq: You can remove isOpen, but can leave the defaultConfig. We may want to add to this later?

Add a space before the equals for defaultConfig=

src/app/modals/about-modal.component.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+    this.prevConfig = cloneDeep(this.config);
+  }
+
+  /**
+   * Close the Modal
+   * @param  $event MouseEvent to emit
+   */
+  close($event: MouseEvent): void {
+    this.onCancel.emit($event);
```
dlabrecq: Please either create a new ModalEvent class or pass true to the emitted event.

A ModalEvent class could contain a close: boolean property.

src/app/modals/about-modal.component.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * The AboutModal config contaning component properties
+   */
+  @Input() config: AboutModalConfig;
+
+  /**
+   * Modal Template for custom content
+   */
+  @Input() template: TemplateRef<any>;
```
dlabrecq: We should probably remove the template and just use ng-content for the body of the modal -- it would be a little cleaner.

src/app/modals/examples/about-modal-example.component.html
```diff
@@ -0,0 +1,41 @@
...
+      <button (click)="openModal(aboutTemplate)" class="btn btn-default">Launch About Modal</button>
+      <ng-template #aboutTemplate>
+        <pfng-about-modal [config]="aboutConfig" 
+                          [template]="aboutModalTemplate" 
+                          (onCancel)="closeModal($event)">
+          <ng-template #aboutModalTemplate>
+            <div>
+              This text is in template.
+            </div>
+          </ng-template>
```
dlabrecq: Lets remove the template and just add HTML directly to the body of the modal.

src/app/modals/about-modal.component.html
```diff
@@ -0,0 +1,25 @@
...
+    <h1 *ngIf="config.title">{{config.title}}</h1>
+    <div *ngIf="config.productInfo && config.productInfo.length > 0" class="product-versions-pf">
+      <ul class="list-unstyled">
+        <li *ngFor="let info of config.productInfo">
+          <strong>{{info.name}}</strong> {{info.value}}</li>
+      </ul>
+    </div>
+    <div class="product-versions-pf">
+      <ng-template *ngIf="template" [ngTemplateOutlet]="template">
+      </ng-template>
```
dlabrecq: Lets use an ng-content tag here.

src/app/modals/examples/about-modal-example.component.html
```diff
@@ -0,0 +1,41 @@
...
+  <div>
+    <tabset>
+      <tab heading="api">
+        <iframe class="demoframe" src="docs/classes/aboutmodalcomponent.html"></iframe>
+      </tab>
+      <tab heading="html">
+        <include-content src="src/app/modals/examples/about-modal-example.component.html"></include-content>
+      </tab>
+      <tab heading="typescript">
+        <include-content src="src/app/modals/examples/about-modal-example.component.ts"></include-content>
```
dlabrecq: When renaming modals to modal, please update these paths

src/app/modals/examples/about-modal-example.component.html
```diff
@@ -0,0 +1,41 @@
...
+      </tab>
+    </tabset>
+  </div>
+</div>
```
dlabrecq: Return char needed

src/app/modals/examples/about-modal-example.component.ts
```diff
@@ -0,0 +1,52 @@
...
+  modalRef: BsModalRef;
+
+  constructor(private modalService: BsModalService) {
+  }
+
+  ngOnInit(): void {
+    this.aboutConfig = {
+      additionalInfo: 'Donec consequat dignissim neque, sed suscipit quam egestas in. Fusce bibendum ' +
+      'laoreet lectus commodo interdum. Vestibulum odio ipsum, tristique et ante vel, iaculis placerat nulla. ' +
+      'Suspendisse iaculis urna feugiat lorem semper, ut iaculis risus tempus.',
```
dlabrecq: Please indent these wrapped lines an additional 2 spaces.

src/app/modals/examples/about-modal-example.component.ts
```diff
@@ -0,0 +1,52 @@
...
+      productInfo: [
+        { name: 'Version', value: '1.0.0.0.20160819142038_51be77c' },
+        { name: 'Server Name', value: 'Localhost' },
+        { name: 'User Name', value: 'admin' },
+        { name: 'User Role', value: 'Administrator' }]
+    } as AboutModalConfig;
+  }
+
+  openModal(template: TemplateRef<any>): void {
+    this.aboutConfig.isOpen = true;
```
dlabrecq: This would have been set by the about modal component, but we don't need it -- please remove.

src/app/modals/examples/about-modal-example.component.ts
```diff
@@ -0,0 +1,52 @@
...
+    this.modalRef = this.modalService.show(template);
+  }
+
+  closeModal($event: MouseEvent): void {
+    this.modalRef.hide();
+    this.aboutConfig.isOpen = false;
```
dlabrecq: Please remove isOpen

src/app/modals/examples/about-modal-example.module.ts
```diff
@@ -0,0 +1,25 @@
...
+
+@NgModule({
+  declarations: [
+    AboutModalExampleComponent
+  ],
+  imports: [
+    CommonModule,
+    DemoComponentsModule,
+    ModalModule.forRoot(),
+    TabsModule.forRoot(),
```
dlabrecq: Please remove TabsModule

src/app/modals/examples/about-modal-example.module.ts
```diff
@@ -0,0 +1,25 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+
+import { ModalModule } from 'ngx-bootstrap/modal';
+import { TabsModule } from 'ngx-bootstrap/tabs';
```
dlabrecq: We're not using tabs to show multiple examples -- please remove

src/app/modals/about-modal-config.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Product copyright information
+   */
+  copyright?: string;
+
+  /**
+   * The alt text for the corner graphic
+   */
+  imgAlt?: string;
```
dlabrecq: Does this apply to the logo in the bottom right corner? If so, logoImageAlt may be more clear. 

Other components are also using 'image' instead of 'img'.

src/app/modals/about-modal-config.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * The source for the corner graphic
+   */
+  imgSrc?: string;
```
dlabrecq: I'm thinking this should be renamed logoImageSrc.

dlabrecq (changes requested):
Looks good overall. Thank you for the PR

dlabrecq:
FYI, the Travis build failed due to a minor TS linter error. Try rebasing with the latest from patternfly-ng and re-build locally.

src/app/modals/examples/about-modal-example.component.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  ngOnInit(): void {
+    this.aboutConfig = {
+      additionalInfo: 'Donec consequat dignissim neque, sed suscipit quam egestas in. Fusce bibendum ' +
+      'laoreet lectus commodo interdum. Vestibulum odio ipsum, tristique et ante vel, iaculis placerat nulla. ' +
+      'Suspendisse iaculis urna feugiat lorem semper, ut iaculis risus tempus.',
+      copyright: 'Trademark and Copyright Information',
+      imgAlt: 'Patternfly Symbol',
+      imgSrc: '',
```
dlabrecq: Try this:

logoImageSrc: '//www.patternfly.org/assets/img/logo-alt.svg',

patternfly-ng.module.ts
sahil143: It's conflicting with ng-bootstrap/modal. Error: Duplicate Identifier

src/app/modals/about-modal.component.html
sahil143: Do we need a `select` attribute for ng-content?

example:
<ng-content [select]="modal-body"></ng-content>

src/app/modals/examples/about-modal-example.module.ts
sahil143: We need TabsModule to show API, html, Typescript docs.

src/app/modals/examples/about-modal-example.module.ts
sahil143: Same as above

src/app/modals/about-modal.component.html
dlabrecq: I don't believe so. That says to add content only if the "element has modal-body attribute".

src/app/modals/examples/about-modal-example.module.ts
dlabrecq: Right. In that case, you'll need to add TabsConfig as well.

index.ts
sahil143: Does this still need to be changed? As, we can't change ModalsModule

patternfly-ng.module.ts
dlabrecq: There isn't a Bootstrap modal in index.js, so I'm assuming the conflict is elsewhere (i.e., with the example)? Regardless, you can simply rename the conflicting import, for example:

import {ModalModule as BSModalModule} from 'ngx-bootstrap/modal';

index.ts
dlabrecq: Please see my previous comment below. You can simply rename the conflicting import, for example:

import {ModalModule as BSModalModule} from 'ngx-bootstrap/modal';

We don't want to use plurals for module and component names.

patternfly-ng.module.ts
sahil143: Conflict is in `about-modal-example.module.ts`. We can import ModalModule as BSModalModule for now but when user uses AboutModal in Angular app, Conflict will probably occur and ModalModule will need to be import as some other name . Is that behaviour ok?

sahil143:
@dlabrecq added some unit tests and other changes. Please review.

dlabrecq (approved):

PR #178, opened 2017-11-04
Included content not displayed

dlabrecq:
Fixed src property for include-content component.

AOT fixes to the example app required a syntax change from "@Input('src')" to "@Input() src: string". As a result, we lost the alias to the templateUrl var -- this fix restores that.


seanforyou23 (approved):
Ship it!

dgutride (approved):

PR #195, opened 2017-11-17
Sort array pipe

dlabrecq:
Created the sort-array pipe, examples, and unit tests.

This work is in preparation for the pin feature of list. It will allow specific items to be sorted and pinned to the top of the list component. 

Also fixed the broken search-highlight pipe example and cleaned up the truncate pipe.

Example:
https://rawgit.com/dlabrecq/patternfly-ng/sort-array-pipe-dist/dist-demo/#/sortArray

### DIFF
--- a/src/app/pipe/truncate/truncate.pipe.ts
+++ b/src/app/pipe/truncate/truncate.pipe.ts
@@ -5,12 +5,17 @@ import { Pipe, PipeTransform } from '@angular/core';
  *
  * This is currently used with the save filter feature of the filter fields component
  */
-@Pipe({ name: 'Truncate'})
+@Pipe({ name: 'truncate'})
 export class TruncatePipe implements PipeTransform {
-  transform(value: string, args: string[]): string {
-    let limit = (args && args.length > 0) ? parseInt(args[0], 10) : 10;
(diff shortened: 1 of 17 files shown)


dlabrecq:
Note that the underlying sort is quicksort, which appears to be performant for our needs. If we must support extremely large data sets, we can swap the algorithm. However, I believe that using the default will be more readable and error resistant.

Looked at various JavaScript sort algorithms (selection sort, insertion sort, bubble sort, heapsort, mergesort, quicksort, the default sort) with 10k items and results are similar. At 250k items, the mergesort and heapsort perform better with heapsort performing best for 500k items or more.

src/app/pipe/truncate/truncate.pipe.ts
```diff
@@ -5,12 +5,17 @@ import { Pipe, PipeTransform } from '@angular/core';
  *
  * This is currently used with the save filter feature of the filter fields component
  */
-@Pipe({ name: 'Truncate'})
+@Pipe({ name: 'truncate'})
```
dgutride, line 8: This will be a breaking change for consumers of this pipe, right?  Can we add support for both Truncate and truncate to help with that?

dgutride (approved):

PR #197, opened 2017-11-20
fix(chart): add missing function and fix typo in chart defaults

jiekang:
Using `getDefaultDonutConfig` fails as one of the functions used is missing and another is named incorrectly. This PR fixes the issue.


dlabrecq (approved):
LGTM. Thank you for the fix!

Issue #180, opened 2017-11-06
Donut chart pattern


catrobson:
Angular implementations:
http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfDonutChart
http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfDonutPctChart

Convert these from Angular 1 to Angular 4.  Have them in our demo application as Donut Chart w/ different implementations (tabs).

jiekang:
I'll try doing this.

catrobson:
Great!  Thanks Jie!

jiekang:
The angular-patternfly donut-pct chart pattern uses pfUtils for standard colors, e.g.
```
      var color = pfUtils.colorPalette.blue;
      [...]
          color = pfUtils.colorPalette.red;
```

What is the correct way to do this in patternfly-ng?

@catrobson Who would be the right person to ask this?

dlabrecq:
@dgutride Created the ground work for the sparkline chart, including chart.defaults.ts. The getDefaultColors property of chart.defaults.ts may have what you need?

FYI, the colorPalette: patternfly.pfPaletteColors of Angular Patternfly is a reference to all PatternFly colors.

jiekang:
@dlabrecq Okay, thanks! For the time being I've added access to `pfPaletteColors` through `chart.defaults.ts` as `getDefaultColors` is only a small subset.

As far as a PR for this issue: I am working on the donut and donut-pct patterns. Would you like these as one PR per pattern? Or a single PR for both?

dlabrecq:
Yes, it is a subset of colors used by the charts.

Regarding PRs, please create one for just the donut chart. Once that has been merged, we can review the donut-pct chart separately.

BTW, it's not clear that donut-pct is short hand for percentage chart. This is an opportunity to clean things up as we port from Angular PatternFly, so perhaps donut-percentage-chart would be a more intuitive name?

jiekang:
Okay. I'll modify my PR.

dlabrecq:
Closing due to inactivity.

PR #200, opened 2017-11-22
Add donut chart pattern

jiekang:
This adds the donut chart pattern addressing one half of issue #180

The patterns closely follow the angularJS versions:
http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfDonutChart

Notable differences:

* similar to the patternfly-ng sparkline pattern, previously stand-alone options like `chartHeight` are moved under the config as for example `config.chartHeight` or `config.centerLabel`

A demo via rawgit is available via:
http://rawgit.com/jiekang/patternfly-ng/donut-chart-dist/dist-demo/#/donut

Thoughts?

### DIFF
--- a/src/app/chart/donut/donut-config.ts
+++ b/src/app/chart/donut/donut-config.ts
@@ -5,9 +5,9 @@ import { ChartConfig } from '../chart-config';
 export class DonutConfig extends ChartConfig {
 
   /**
-   * An optional function to customize the text of the center label
+   * Text for the donut chart center label (optional)
    */
-  centerLabelFn?: any;
+  centerLabel?: any;
 
   /**
    * An optional function to handle when donut is clicked
@@ -15,12 +15,18 @@ export class DonutConfig extends ChartConfig {
   onClickFn?: any;
 
   /**
-   * The height of the donut chart
+   * The height of the donut chart (optional)
    */
   chartHeight?: number;
 
   /**
    * C3 inherited configuration for colors
+   * An object with key-value pairs of data name and color, e.g.
+   * colors : {
+   *   Cats: '#0088ce',
+   *   Hamsters: '#3f9c35',
+   * }
+   *
    */
   colors?: any;
 
--- a/src/app/chart/donut/donut-config.ts
+++ b/src/app/chart/donut/donut-config.ts
@@ -1,0 +1,36 @@
+import { ChartConfig } from '../chart-config';
+/**
+ * A config containing properties for the sparkline chart
+ */
+export class DonutConfig extends ChartConfig {
+
+  /**
+   * An optional function to customize the text of the center label
+   */
+  centerLabelFn?: any;
+
+  /**
+   * An optional function to handle when donut is clicked
+   */
+  onClickFn?: any;
+
+  /**
+   * The height of the donut chart
+   */
+  chartHeight?: number;
+
+  /**
+   * C3 inherited configuration for colors
+   */
+  colors?: any;
+
+  /**
+   * C3 inherited configuration for size
+   */
+  size?: any;
+
+  /**
+   * C3 inherited donut configuration
+   */
+  donut?: any;
+}
--- a/src/app/chart/donut/donut-config.ts
+++ b/src/app/chart/donut/donut-config.ts
@@ -39,4 +39,9 @@ export class DonutConfig extends ChartConfig {
    * C3 inherited donut configuration
    */
   donut?: any;
+
+  /**
+   * C3 inherited configuration for tooltip
+   */
+  tooltip?: any;
 }
--- a/src/app/chart/donut/donut.component.html
+++ b/src/app/chart/donut/donut.component.html
@@ -1,0 +1,1 @@
+<div #chartElement id="{{donutChartId}}"></div>
--- a/src/app/chart/donut/donut.component.html
+++ b/src/app/chart/donut/donut.component.html
@@ -1,1 +1,1 @@
-<div #chartElement id="{{donutChartId}}"></div>
+<div #chartElement id="{{config.chartId}}"></div>
--- a/src/app/chart/donut/donut.component.spec.ts
+++ b/src/app/chart/donut/donut.component.spec.ts
@@ -20,11 +20,13 @@ describe('Component: donut chart', () => {
 
   beforeEach(() => {
     config = {
-      'chartId': 'testDonutChart',
-      data: {},
+      chartId: 'testDonutChart',
       onClickFn: function(d: any, e: any) {
       },
-      centerLabel: 'center'
+      data: {},
+      donut: {
+        title: 'Animals'
+      }
     };
     data = [
       ['Cats', 2],
@@ -50,6 +52,10 @@ describe('Component: donut chart', () => {
       });
   }));
 
+  it('should set chart id', () => {
+    expect(comp.donutChartId).toBe('testDonutChartdonutChart');
+  });
+
 
   it('should allow attribute specification of chart height', () => {
     config.chartHeight = 120;
@@ -61,14 +67,12 @@ describe('Component: donut chart', () => {
     config.chartHeight = 120;
 
     fixture.detectChanges();
-    expect(comp.donutChartId).toBe('testDonutChartdonutChart');
     expect(comp.config.size.height).toBe(120);
 
     config.chartHeight = 100;
     fixture.detectChanges();
     expect(comp.config.size.height).toBe(100);
   });
-
 
   it('should setup C3 chart data correctly', () => {
     expect(comp.config.data.columns.length).toBe(3);
@@ -89,5 +93,31 @@ describe('Component: donut chart', () => {
 
   it('should setup onclick correctly', () => {
     expect(typeof(comp.config.data.onclick)).toBe('function');
+  });
+
+  it('should use the default centerLabel', () => {
+    let centerLabel = comp.getCenterLabelText();
+    expect(centerLabel.bigText).toBe(6);
+    expect(centerLabel.smText).toBe('Animals');
+  });
+
+  it('should use custom centerLabel', () => {
+    config.centerLabel = 'custom-label';
+    fixture.detectChanges();
+
+    let centerLabel = comp.getCenterLabelText();
+    expect(centerLabel.bigText).toBe('custom-label');
+    expect(centerLabel.smText).toBe('');
+  });
+
+  it('should use patternfly tooltip', () => {
+    expect(typeof(comp.config.tooltip.contents)).toBe('function');
+  });
+
+  it('should have default donut config with custom title', () => {
+    expect(comp.config.donut.title).toBe('Animals');
+
+    expect(comp.config.donut.width).toBe(11);
+    expect(comp.config.donut.label.show).toBe(false);
   });
 });
--- a/src/app/chart/donut/donut.component.spec.ts
+++ b/src/app/chart/donut/donut.component.spec.ts
@@ -53,7 +53,7 @@ describe('Component: donut chart', () => {
   }));
 
   it('should set chart id', () => {
-    expect(comp.donutChartId).toBe('testDonutChartdonutChart');
+    expect(comp.config.chartId).toContain('testDonutChart');
   });
 
 
--- a/src/app/chart/donut/donut.component.spec.ts
+++ b/src/app/chart/donut/donut.component.spec.ts
@@ -24,8 +24,7 @@ describe('Component: donut chart', () => {
       data: {},
       onClickFn: function(d: any, e: any) {
       },
-      centerLabelFn: function() {
-      }
+      centerLabel: 'center'
     };
     data = [
       ['Cats', 2],
--- a/src/app/chart/donut/donut.component.spec.ts
+++ b/src/app/chart/donut/donut.component.spec.ts
@@ -1,0 +1,94 @@
+import {
+  async,
+  ComponentFixture,
+  TestBed
+} from '@angular/core/testing';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { FormsModule } from '@angular/forms';
+import { By } from '@angular/platform-browser';
+import { DonutConfig } from './donut-config';
+import { DonutComponent } from './donut.component';
+import { ChartDefaults } from '../chart.defaults';
+
+describe('Component: donut chart', () => {
+
+  let comp: DonutComponent;
+  let fixture: ComponentFixture<DonutComponent>;
+
+  let config: DonutConfig;
+  let data: any;
+
+  beforeEach(() => {
+    config = {
+      'chartId': 'testDonutChart',
+      data: {},
+      onClickFn: function(d: any, e: any) {
+      },
+      centerLabelFn: function() {
+      }
+    };
+    data = [
+      ['Cats', 2],
+      ['Hamsters', 2],
+      ['Dogs', 2]
+    ];
+  });
+
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      imports: [BrowserAnimationsModule, FormsModule],
+      declarations: [DonutComponent],
+      providers: [ChartDefaults]
+    })
+      .compileComponents()
+      .then(() => {
+        fixture = TestBed.createComponent(DonutComponent);
+        comp = fixture.componentInstance;
+        comp.config = config;
+        comp.chartData = data;
+        fixture.detectChanges();
+      });
+  }));
+
+
+  it('should allow attribute specification of chart height', () => {
+    config.chartHeight = 120;
+    fixture.detectChanges();
+    expect(comp.config.size.height).toBe(120);
+  });
+
+  it('should update when the chart height attribute changes', () => {
+    config.chartHeight = 120;
+
+    fixture.detectChanges();
+    expect(comp.donutChartId).toBe('testDonutChartdonutChart');
+    expect(comp.config.size.height).toBe(120);
+
+    config.chartHeight = 100;
+    fixture.detectChanges();
+    expect(comp.config.size.height).toBe(100);
+  });
+
+
+  it('should setup C3 chart data correctly', () => {
+    expect(comp.config.data.columns.length).toBe(3);
+    expect(comp.config.data.columns[0][0]).toBe('Cats');
+    expect(comp.config.data.columns[1][0]).toBe('Hamsters');
+  });
+
+  it('should update C3 chart data when data changes', () => {
+    expect(comp.config.data.columns.length).toBe(3);
+    expect(comp.config.data.columns[0][0]).toBe('Cats');
+    expect(comp.config.data.columns[0][1]).toBe(2);
+
+    data[0][1] = 3;
+    fixture.detectChanges();
+
+    expect(comp.config.data.columns[0][1]).toBe(3);
+  });
+
+  it('should setup onclick correctly', () => {
+    expect(typeof(comp.config.data.onclick)).toBe('function');
+  });
+});
--- a/src/app/chart/donut/donut.component.ts
+++ b/src/app/chart/donut/donut.component.ts
@@ -1,6 +1,6 @@
 import { Component, DoCheck, Input, OnInit } from '@angular/core';
 
-import { cloneDeep, defaults, isEqual, merge } from 'lodash';
+import { cloneDeep, defaults, isEqual, merge, uniqueId } from 'lodash';
 import { Subscription } from 'rxjs/Subscription';
 
 import { ChartDefaults } from '../chart.defaults';
@@ -18,8 +18,6 @@ export class DonutComponent extends ChartBase implements DoCheck, OnInit {
   @Input() chartData: any;
   @Input() config: DonutConfig;
 
-  public donutChartId: any;
-
   private prevChartData: any;
 
   private subscriptions: Subscription[] = [];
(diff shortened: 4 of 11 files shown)


joshuawilson:
I think a lot of those commits can/should be squashed. A commit should be a change, if it takes you several commits to get it right but they are all the same thing then condense them. It keeps the change atomic.

jiekang:
Okay. I've squashed the commits and following comments in #180 only included the donut chart pattern for this PR.

dlabrecq:
Ideally, these components should be separate PRs -- it's a bit much to review at once. 

If there are two components, they should have their own examples. Tabs can be used to show variations of that example. 

Please include a link to a working example. A screenshot is also very helpful. I can download and build your code, but others (designers) will need to review.

There are instructions in the contribution guide for how to setup Travis to generate an example for you.

joshuawilson:
From my experience that is not true. If you are on 2.4.5 and commit 10 fixes and 1 feat you will be on 2.5.0 not 2.5.10.
And for fixes be it 1 or 100, it only increments 1.

dlabrecq:
Hmm, good to know. I thought we had already seen that happen with patternfly-ng? I'll just have to keep an eye on the semantic release. 

Still seems like we would only want one "feat" commit per PR regardless of how many fixes are applied after.

joshuawilson:
I think a PR should be small, so I agree. But it is not the PR that does the release. Even merging to master may fail in the build and not get released. There have been times when the PRs merged but no release went out, then a giant stack of commits formed a new release.

dlabrecq:
Actually, that's exactly what I was seeing. It's odd to have so many fixes in a new feature being contributed for the first time? I understand that's how it works, but seems like that should just be 2.5.0, not 2.5.10 for a new feature. .Anyway, I probably shouldn't hijack this PR for this topic.

jiekang:
I'd like to setup Travis to generate an example for me. The contribution guide states:

_You will also need to add an AUTH_TOKEN variable to Travis generated in your GitHub account to allow Travis to connect to your fork._


Do you have more detailed instructions on how to accomplish this? My Travis build fails as it's not authorized to push to my repository.

In the meantime I've attached a link to a screenshot of the donut example page:

[Screenshot](https://pasteboard.co/GUSytG3.png)

sahil143:
@jiekang 
[Travis Build](https://github.com/patternfly-webcomponents/patternfly-webcomponents/blob/master/TESTS.md#enabling-travis-builds)
You can follow this step by step guide to add AUTH_TOKEN to Travis. Remember, you need to do this for Patternfly-ng instead of Patternfly-webcomponents.

jiekang:
@sahil143 Thank you!

jiekang:
A demo via rawgit is available at:

http://rawgit.com/jiekang/patternfly-ng/donut-chart-dist/dist-demo/#/

The donut chart is routed at:

http://rawgit.com/jiekang/patternfly-ng/donut-chart-dist/dist-demo/#/donut

dlabrecq:
Does the donut chart provide the ability to add horizontal and vertical labels like the Angular PatternFly example?

jiekang:
@dlabrecq 
I don't see any mention of horizontal or vertical labels in the API [1] or source [2] for donut charts.
[1] http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfDonutChart
[2] https://github.com/patternfly/angular-patternfly/blob/master/src/charts/donut/donut-chart-component.js

For context, the code is as close to a 1 to 1 transform of `js` to `ts` and `angularjs` to `angular` that tries to follow the style of the sparkline chart that's already in patternfly-ng. The only major differences are:

* Does not have `dataAvailable` html guard to switch display to `pf-empty-chart`. I don't think this element exists yet in patternfly-ng
* Uses the `EventEmitter` provided by `ChartBase` to access `chart` when it's available. The angularjs version accesses `chart` directly.
* Merges defaults on `config` and `config.donut` as the `defaults` function is not recursive and `donut` contains default configuration that we'd like to keep. Otherwise a user setting the title via `donut.title` will lose all default configuration in `config.donut`. In `angularjs` it seems the merge is recursive as this problem isn't seen with the exact same example.

In general, there should be no functional difference between the `angular` and `angularjs` versions of the donut chart. At least that's what I intended.

I see a labelConfig in the donut-pct API [3]. Is that what you are referring to? It is not available in the donut chart.
[3] http://www.patternfly.org/angular-patternfly/#/api/patternfly.charts.component:pfDonutPctChart

dlabrecq:
The labels I'm referring to in the Angular PatternFly donut chart example below are; Cats, Hamsters, Fish, and Dogs.

screen shot 2017-11-25 at 12 03 13 am

jiekang:
Ah, yes users can add a legend. In general, users can add c3 configuration in the same manner as before (which is how the legend is added in the angularjs example).

@dlabrecq Would you like me to modify the angular example to include exactly the two that you showed?

dlabrecq:
Yes, please do. Need to ensure the examples can demonstrate feature parity with APF.

jiekang:
@dlabrecq Okay, done. 

Looks like the rawgit has automatically updated and the result can be viewed there.

src/app/chart/donut/donut-config.ts
```diff
@@ -0,0 +1,36 @@
+import { ChartConfig } from '../chart-config';
+/**
+ * A config containing properties for the sparkline chart
+ */
+export class DonutConfig extends ChartConfig {
+
+  /**
+   * An optional function to customize the text of the center label
+   */
+  centerLabelFn?: any;
```
dlabrecq: In general, we want to stay away from callback functions. Instead, this can be a simple centerLabel property.

src/app/chart/donut/donut-config.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * An optional function to handle when donut is clicked
+   */
+  onClickFn?: any;
```
dlabrecq, line 15: Instead of a callback function, we would like to have an onClick event in the component.

src/app/chart/donut/donut-config.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * The height of the donut chart
+   */
+  chartHeight?: number;
+
+  /**
+   * C3 inherited configuration for colors
+   */
+  colors?: any;
```
dlabrecq: It's not clear what "colors" is. Please add an example to the comments? For example:

'colors' : {
&nbsp;&nbsp;'Cats': '#0088ce',     // blue
&nbsp;&nbsp;'Hamsters': '#3f9c35', // green
&nbsp;&nbsp;'Fish': '#ec7a08',     // orange
&nbsp;&nbsp;'Dogs': '#cc0000'      // red
}

I wonder if a typed object may help here?

colors?: Color;

src/app/chart/donut/donut.component.html
```diff
@@ -0,0 +1 @@
+<div #chartElement id="{{donutChartId}}"></div>
```
dlabrecq: We will need to create the pf-empty-chart for the no data available use case. This is an integral part of the component HTML layout in Angular Patternfly. It should be a very simple component -- please let me know if you need help with that?

src/app/chart/donut/donut.component.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    data[0][1] = 3;
+    fixture.detectChanges();
+
+    expect(comp.config.data.columns[0][1]).toBe(3);
+  });
+
+  it('should setup onclick correctly', () => {
+    expect(typeof(comp.config.data.onclick)).toBe('function');
+  });
+});
```
dlabrecq: These tests are a great start.

In addition to these tests, I would like to see a couple tests ported from donut-pct.spec.js. For example, anything that tests the label would be very useful here. The donut-pct tests seem to share a lot in common with donut.

src/app/chart/donut/donut.component.ts
```diff
@@ -0,0 +1,144 @@
...
+@Component({
+  selector: 'pfng-chart-donut',
+  templateUrl: './donut.component.html'
+})
+export class DonutComponent extends ChartBase implements DoCheck, OnInit {
+
+  @Input() chartData: any;
+  @Input() config: DonutConfig;
+
+  public donutChartId: any;
```
dlabrecq: I'm thinking this should be renamed as "id". However, what is the difference between donutChartId and config.chartId? 

We don't have an id attribute for any other patternfly-ng component, so can this just be removed?

src/app/chart/donut/donut.component.ts
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * Default constructor
+   * @param chartDefaults
+   */
+  constructor(private chartDefaults: ChartDefaults) {
+    super();
+  }
+
+  ngOnInit(): void {
+    this.donutChartId = 'donutChart';
```
dlabrecq: There could be more than one chart on a page, so assigning the same ID won't scale well. 

I realize this is a direct port of APF, but this looks like a bug to me. I'm thinking we append some sort of auto-generated number or just leave it blank?

If the chart needs an id to render, then we can throw an exception, forcing the dev to set one?

src/app/chart/donut/donut.component.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (this.config.chartId) {
+      this.donutChartId = this.config.chartId + this.donutChartId;
+    }
+
+    this.chartLoaded.subscribe({
+      next: (chart: any) => {
+        this.chartAvailable(chart);
+      }
+    });
```
dlabrecq: If we subscribe, we also need to unsubscribe when the component is destroyed. Otherwise, it will ultimately contribute to memory leaks as we discovered in OSIO.

In OSIO, we created an array, named subscriptions. The code looks something like this:

subscriptions: Subscription[] = [];
...
constructor() {
    this.subscriptions.push(this.chartLoaded.subscribe...
...

 ngOnDestroy(): void {
    this.subscriptions.forEach(sub => {
      sub.unsubscribe();
    });
  }

src/app/chart/donut/donut.component.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    this.setupConfig();
+    super.generateChart(this.donutChartId);
+    this.updateAll();
```
dlabrecq: You appear to be calling updateAll and generateChart twice. Once via ngDoCheck and again here. 

I believe we want to call generateChart and updateAll once, within ngDoCheck/setupConfig? That is, if and only if there is a change to chartData or config.

src/app/chart/donut/donut.component.ts
```diff
@@ -0,0 +1,144 @@
...
+      this.updateAll();
+    }
+
+    if (!isEqual(this.config, this.prevConfig)) {
+      this.setupConfig();
+      this.generateChart(this.donutChartId, true);
+    }
+  }
+
+  public chartAvailable(chart: any): void {
```
dlabrecq: Can this be private?

src/app/chart/donut/donut.component.ts
```diff
@@ -0,0 +1,144 @@
...
+      columns: this.chartData,
+      order: null,
+      colors: this.config.colors
+    };
+  }
+
+  private updateAll(): void {
+    this.prevChartData = cloneDeep(this.chartData);
+
+    this.config.data = merge(this.config.data, this.getDonutData(this.chartData));
```
dlabrecq: If you're merging data, the config should be cloned after this change. That is, ensure setupConfig is always called after updateAll

src/app/chart/donut/examples/donut-example.module.ts
```diff
@@ -0,0 +1,23 @@
...
+import { DonutExampleComponent } from './donut-example.component';
+import { ChartModule } from '../../chart.module';
+
+@NgModule({
+  declarations: [DonutExampleComponent],
+  imports: [
+    CommonModule,
+    FormsModule,
+    DemoComponentsModule,
+    ChartModule,
```
dlabrecq: Alphabetical order please

src/demo/app-routing.module.ts
```diff
@@ -20,6 +20,7 @@ import { SearchHighlightExampleComponent }
 import { SortExampleComponent } from '../app/sort/examples/sort-example.component';
 import { SortArrayExampleComponent } from '../app/pipe/sort-array/examples/sort-array-example.component';
 import { SparklineExampleComponent } from '../app/chart/sparkline/examples/sparkline-example.component';
+import { DonutExampleComponent } from '../app/chart/donut/examples/donut-example.component';
```
dlabrecq: Alphabetical order please

src/demo/app-routing.module.ts
```diff
@@ -85,6 +86,9 @@ const routes: Routes = [{
   }, {
     path: 'sparkline',
     component: SparklineExampleComponent
+  }, {
+    path: 'donut',
```
dlabrecq: Alphabetical order please

src/demo/app.component.html
```diff
@@ -37,6 +37,9 @@ <h3>Components</h3>
             <li>
               <a routerLink="/sparkline" routerLinkActive="active">Sparkline</a>
             </li>
+            <li>
+              <a routerLink="/donut" routerLinkActive="active">Donut</a>
```
dlabrecq: Alphabetical order please

src/demo/app.module.ts
```diff
@@ -55,6 +56,7 @@ import { NavigationExampleModule } from '../app/navigation/examples/navigation-e
     SortArrayExampleModule,
     SortExampleModule,
     SparklineExampleModule,
+    DonutExampleModule,
```
dlabrecq: Alphabetical order please

src/demo/app.module.ts
```diff
@@ -27,6 +27,7 @@ import { SearchHighlightExampleModule } from '../app/pipe/search-highlight/examp
 import { SortArrayExampleModule } from '../app/pipe/sort-array/examples/sort-array-example.module';
 import { SortExampleModule } from '../app/sort/examples/sort-example.module';
 import { SparklineExampleModule } from '../app/chart/sparkline/examples/sparkline-example.module';
+import { DonutExampleModule } from '../app/chart/donut/examples/donut-example.module';
```
dlabrecq: Alphabetical order please

dlabrecq (changes requested):
Please update the example to show tooltips like Angular PatternFly.

src/app/chart/donut/donut.component.ts
jiekang: `config.chartId` is user specified. `donutChartId` is generated from this. This section is actually an interpretation of the patternfly-ng code in `sparkline.component.ts` rather than from angular-patternfly, pasted below for reference:

```
    // Create an ID for the chart based on the chartId in the config if given
    if (this.sparklineChartId === undefined) {
      this.sparklineChartId = 'sparklineChart';
      if (this.config.chartId) {
        this.sparklineChartId = this.config.chartId + this.sparklineChartId;
      }
    }
```

The undefined check suggests users can input sparklineChartId themselves but the API and all the examples in angular-patternfly and patternfly-ng direct users to modify config.chartId. My interpretation drops this check, possibly incorrectly. Maybe there was an intention for users to set these 'new' chart ids in patternfly-ng? I don't know the history here, sorry.

src/app/chart/donut/donut.component.ts
jiekang: Right. I like the idea of auto-generating a uuid to attach behind the scenes. When using the sparkline chart in f8-ui I ran into the chart id issue where id's on the same page need to be unique (whereas for some other things, the id's only need to be unique for the same angular template). So multiple templates on the same page could conflict if the id's happened to be the same for the charts. It would be nice to have an auto-generated uniqueness behind the scenes.

src/demo/app-routing.module.ts
jiekang: Done.

Thoughts on adding `ordered-imports` and the like to enforce tslinting of alphabetical order as a separate issue/PR?  Also I noticed `tslint` was updated from 4.x to 5.x and the tslint config file contains unsupported configuration for `member-ordering`. It doesn't seem to break anything but I think a fix would be good for those using editors that read from the config file to show tslint warnings.

src/app/chart/donut/donut.component.ts
jiekang: Nice catch. Done.

src/app/chart/donut/donut.component.html
jiekang: I presume this is another issue and PR? I can take a look at porting this. I suppose it's:

https://github.com/patternfly/angular-patternfly/blob/master/src/charts/empty-chart.component.js
https://github.com/patternfly/angular-patternfly/blob/master/src/charts/empty-chart.html

src/app/chart/donut/donut.component.ts
jiekang: Right, done.

src/app/chart/donut/donut.component.ts
jiekang: This comment and the one on `updateAll` triggered a re-evaluation of the update code. For context this was originally taken from the `sparkline` component already in patternfly-ng. I have revised the design heavily and I think the original in `sparkline` component could also be revised in a similar matter. I have also included a temporary commit in the branch to demonstrate the update capabilities for both the data-only and config-only situation via the examples. Please let me know what you think.

In general I think there is still an issue that makes working with the code more difficult than necessary, however I did not want to make any changes that would affect all charts in this PR.

This is that the user inputs `config` and `data` separately in the component, while internally, these are combined under the `config` object (`config && config.data`). This ends up making updates more complicated as an update to `config` should trigger a full reload of the chart, while an update of the users `data` object only needs to trigger a reload of the data. Since they're combined, we cannot make `config.data` a reference to the users data object, as changes to the users data would result in config being not equal, triggering a full reload instead of just a data reload. At the moment, to workaround this, it's a deep clone instead (from the `merge` function), resulting in extra memory, and extra cycles used.

src/app/chart/donut/donut-config.ts
jiekang: This onClickFn mirrors `angular-patternfly` and would be an API difference between `angular-patternfly` and `patternfly-ng` if changed, is that okay? My approach was to keep things one to one such that anyone deciding to port from one to the other would not have to make any config changes due to incompatibilities. However I realize `angularjs` and `angular` are completely different projects so this might not be a concern for this project.

I don't understand what an "onClick event in the component" means, could you elaborate?

src/app/chart/donut/donut-config.ts
jiekang: Changed to `centerLabel` property.

src/app/chart/donut/donut-config.ts
jiekang: I've updated the comment with an example. I'm reluctant to add a typed object for c3 inherited configuration as it mimics c3 configuration exactly. Maybe a link to the c3 API that it's intended to match would work for all c3 inherited configuration?

src/app/chart/donut/donut.component.ts
dlabrecq: Having two IDs is confusing and don't see the need for it. I'd like to clean this up for both the sparkline and donut charts and just use config.chartId.

src/demo/app-routing.module.ts
dlabrecq: Yes, it would be good to fix this. A separate PR would probably be best.

dlabrecq: The member-ordering seems to be working; for example, if I add a private method before a public one. Although, it's possible that config could be deprecated?

Unfortunately, none of that enforces import ordering, just fields, methods, and constructors. However, I did find adding this does the trick:

"ordered-imports": true,

src/app/chart/donut/donut.component.ts
dlabrecq: It appears that merge with config.data was taken almost verbatim from APF. It was using a merge utility, but no idea why. Regardless, we're not using that and the chart doesn't appear to need it. Lets get rid of that and keep the data and config separate.

src/app/chart/donut/donut-config.ts
dlabrecq: Yes, it's ok to be different from APF. This is our opportunity to clean things up.

Instead of calling a function when the donut is clicked, we emit an event. A developer can easily subscript to the event like so:

<pfng-donut-chart (onClick)="myFunction($event)"...

Then, in the component we would have something like this:

/**
&nbsp;* The event emitted when an item has been clicked
&nbsp;*/
@Output('onClick') onClick = new EventEmitter();

And some method to handle the click:

handleClick($event: MouseEvent): void {
&nbsp;&nbsp;this.onClick.emit({
&nbsp;&nbsp;&nbsp;&nbsp;item: item
&nbsp;&nbsp;} as ListEvent);
}

src/app/chart/donut/examples/donut-example.component.ts
```diff
@@ -54,5 +55,15 @@ export class DonutExampleComponent implements OnInit {
...
+
+    Observable
+      .timer(0, 1000)
+      .map(() => Math.floor(Math.random() * 5) + 1)
+      .subscribe(val => this.data[0][1] = val);
+
+    Observable
+    .timer(0, 5000)
+    .map(() => Math.floor(Math.random() * 100) + 100)
+    .subscribe(val => this.config2.chartHeight = val);
```
dlabrecq, line 67: This is very cool, but can we add this as a separate example? For example, have a "Basic" and "Dynamic" tabs similar to other component examples.

Sometimes my tooltip disappears as I'm attempting to hover and the data changes under me.

dlabrecq: As an example, we could have examples named donut-basic-example.component.ts and donut-dynamic-example.component.ts...

dlabrecq (approved):
LGTM

PR #201, opened 2017-11-25
Support for Angular 4 and 5

dlabrecq:
Replaced dependencies with peerDependencies in order for patternfly-ng to work with both Angular 4 and Angular 5 apps. By using peer dependencies, it allows the app to generate the factories -- patternfly-ng only provides the *.metadata.json files. 

Also cleaned up several, unused dev dependencies.

Tested AOT with Angular 4 and Angular 5, minus the tree list component.

Note that angular-tree-component still uses deprecated ngOutletContext, which causes Angular 5 builds to fail. See: https://github.com/500tech/angular-tree-component/issues/483


adamkleingit:
angular-tree-component version 5.2.0 uses the new ngTemplateOutletContext

dlabrecq:
angular-tree-component v6.0.0 builds fine, so this should be all set.

dgutride (approved):

PR #214, opened 2017-12-02
Sparkline and donut chart clean up

dlabrecq:
Refactored sparkline and donut charts to update config and chart data properly. Whenever chartData is merged with config.data, the component would detect an unnecessary change.

Also fixed:

- Fixed merging of chartData to c3 config.data
- Moved dynamic chart to its own example tab
- Consolidated multiple ID properties (id, bindto, and chartId) as chartId
- Added unique id if chartId is not provided
- Added missing exports to chart.module and index.js
- Added missing API comments for TypeDoc
- Added missing config and data types
- Added missing window ref
- Created SparklineData type
- Destroy subscriptions

https://rawgit.com/dlabrecq/patternfly-ng/chart-dist/dist-demo/#/donut

### DIFF
--- a/src/app/card/basic-card/examples/card-trend-example.component.ts
+++ b/src/app/card/basic-card/examples/card-trend-example.component.ts
@@ -7,6 +7,8 @@ import {
 import { CardAction } from '../../card-action/card-action';
 import { CardConfig } from '../card-config';
 import { CardFilter } from '../../card-filter/card-filter';
+import { SparklineConfig } from '../../../chart/sparkline/sparkline-config';
+import { SparklineData } from '../../../chart/sparkline/sparkline-data';
 
 @Component({
   encapsulation: ViewEncapsulation.None,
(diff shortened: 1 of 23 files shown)


src/app/card/basic-card/examples/card-trend-example.component.ts
```diff
@@ -16,38 +18,38 @@ import { CardFilter } from '../../card-filter/card-filter';
...
-  chartDataVirtual: any = {
+  chartDataVirtual: SparklineData = {
     dataAvailable: true,
     total: 250,
     xData: this.chartDates,
     yData: [
       'used', '90', '20', '30', '20', '20', '10', '14', '20', '25',
       '68', '44', '56', '78', '56', '67', '88', '76', '65', '87', '76']
   };
-  chartConfigPhysical: any = {
```
dgutride, line 32: seems odd to use a sparkline config/data in here instead of something related to a trend view

dgutride (approved):

src/app/card/basic-card/examples/card-trend-example.component.ts
dlabrecq: I believe this is just temporary because we don't have any other chart to show right now.

Issue #222, opened 2017-12-06
Add support for Observables in Notification Service


seanforyou23:
Currently, there is an issue with the NotificationService where, if an app uses the `OnPush` ChangeDetectionStrategy, notifications are not immediately displayed after being added with notificationService.message().

Providing observables as a means of getting/setting notifications is one way of solving this issue.

Two separate upstream applications have found the need for observables with regard to notifications

https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/shared/notifications.service.ts
and
https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/common/ui-patternfly/notification-service.ts

So, this issue is for investigating and providing support for observables to allow for a more robust interface for interacting with notifications. This may require augmenting the existing notification service or creating a new service altogether.

dlabrecq:
This is available now

PR #231, opened 2017-12-12
feat(notifications): add observable for watching notifications array

seanforyou23:
This PR introduces a new property (getNotificationsObserver) to the notification service which wraps the notifications array in an observable. This became necessary when we realized the existing getNotifications() method doesn't work so well while using the OnPush ChangeDetectionStrategy in Angular.

Solves [[issue-222]](https://github.com/patternfly/patternfly-ng/issues/222)

One specific thing I'd like feedback on is how I'm using clearInterval() for the example timer. It works fine in general, but when switching between the tabs quickly you end up building a queue of notifications because the clearInterval doesn't seem to get called if that view is not active. A better approach may be to clearInterval on switching between the tabs, but how to access the timer from within the tabSelected event in notification-service-example.component.ts?

https://rawgit.com/seanforyou23/patternfly-ng/notifications-observable-dist/dist-demo/#/notificationservice

add a new getNotificationsObserver property which allows for listening in on notifications activity
update stream on add/remove notification from data source
add a simple wrapper function for updating the stream to enforce DRY
includes example files for demo app

### DIFF
--- a/src/app/notification/examples/notification-service-observer-example.component.ts
+++ b/src/app/notification/examples/notification-service-observer-example.component.ts
@@ -20,32 +20,28 @@ import { Observable } from 'rxjs';
     notifications: Observable<Notification[]>;
 
     constructor(private notificationService: NotificationService) {
-      notificationService.setDelay(8000);
     }
 
     ngOnInit(): void {
       this.notifications = this.notificationService.getNotificationsObserver;
       // Display notifications to illustrate how
       // one can react to a notifications observer
-      const exampleTimer = setInterval(_ => {
-        this.notificationService.message(
-          NotificationType.SUCCESS,
-          'Example header',
-          'Open js console to see notifications observer in action',
-          false,
-          null,
-          null
-        );
-      }, 1000);
-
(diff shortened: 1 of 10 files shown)


src/app/notification/examples/notification-service-observer-example.component.ts
```diff
@@ -0,0 +1,51 @@
...
+          'Open js console to see notifications observer in action',
+          false,
+          null,
+          null
+        );
+      }, 1000);
+
+      setTimeout(_ => {
+        clearInterval(exampleTimer);
+      }, 6000);
```
dlabrecq: You might consider replacing the timer with a button? It may be a cleaner way to demo this example without a building up messages while switching tabs. At least you won't generate more messages than you can click.

On another note, the basic and observer examples should be able to use the same notification service. Even if there are a build up of messages, I expect the basic example would still show those messages, right?

dlabrecq (changes requested):
Can you please provide a link to your example?

dlabrecq (approved):
LGTM. Great job!

Issue #223, opened 2017-12-07
Demo menu items hidden from view


seanforyou23:
On the [demo site](http://www.patternfly.org/patternfly-ng/) there is a styling issue where the left hand nav is cut short, hiding the Services section. This makes it difficult for users to find information related to the [NotificationService](http://www.patternfly.org/patternfly-ng/#/notificationservice).

Attached image reveals what's missing from view.

<img width="222" alt="screen shot 2017-12-06 at 8 10 41 pm" src="https://user-images.githubusercontent.com/5942899/33693717-609748f2-dac2-11e7-9652-bec0e4947da7.png">

dlabrecq:
It might be time to introduce the vertical nav component here? (Now that we have more components to list) Unless, you have a quick fix in mind?

seanforyou23:
I don't see a quick fix that's low touch, either way it seems like the vertical nav is the way to go. So, introducing the vertical nav has my vote.

Should I create a new issue for that or should I use this ticket as the basis for that work?

dlabrecq:
Please stand by. I have something in the works.

dlabrecq:
Fixed by: https://github.com/patternfly/patternfly-ng/pull/232

PR #232, opened 2017-12-14
Demo: navigation layout based on patternfly.org

dlabrecq:
Added a new navigation layout for the demo based on patternfly.org.

Note that this layout is responsive. When the window shrinks, the primary nav is replaced with a hamburger icon. The user can then navigate all primary menu items via the side bar.

Fixes: https://github.com/patternfly/patternfly-ng/issues/223

Example:
https://rawgit.com/dlabrecq/patternfly-ng/navbar-dist/dist-demo/


dgutride (dismissed):

dgutride:
I like it - only thought was I couldn't find the components at first, might be good to get one of the designers to weigh in on the layout, too?

AdamJ:
Would it make sense to have the structure of the vertical navigation match (or closely resemble) PatternFly.org, to keep some continuity between the two?

dlabrecq:
I didn't follow patternfly.org verbatium because its categories make it very difficult to find components. However, I've changed things up a bit. You can now navigate anywhere from Get Started, which is also the landing page. 

In addition, I added selection highlighting which seems to help a lot.

Get Started:
screen shot 2017-12-14 at 11 43 38 am

Components:
screen shot 2017-12-14 at 11 43 57 am

seanforyou23:
Looks great! One thing I noticed is the menu doesn't update itself when the browser back button is used. Similarly, if I navigate to a component using the menu and then click the logo in the top left corner, the menu doesn't reflect that we're now on the home view. It does however update itself when clicking the heading (.secondary-menu-toggle) at the top of the menu, though the cursor isn't a pointer so it isn't so obvious.

dlabrecq:
Thanks @seanforyou23. The browser back button never worked for the existing demo -- the selected item in the navigation pane never changes. Regardless, I've added support for the browser back button and it seems to work well now.

I've also spoken to @mindreeper2420 offline. Per his comments, I've removed the 200px, left-side margin to fix the layout on mobile.

AdamJ (approved):

PR #256, opened 2018-01-08
fix(toolbar): add support to load queries on filter field select.

vikram-raj:
Fix: https://github.com/fabric8-ui/fabric8-planner/issues/2389

### DIFF
--- a/src/app/filter/filter-fields.component.ts
+++ b/src/app/filter/filter-fields.component.ts
@@ -14,6 +14,7 @@ import { FilterConfig } from './filter-config';
 import { FilterEvent } from './filter-event';
 import { FilterField } from './filter-field';
 import { FilterQuery } from './filter-query';
+import { FilterType } from './filter-type';
 
 /**
  * Component for the filter query field and filter query dropdown
@@ -219,10 +220,10 @@ export class FilterFieldsComponent implements DoCheck, OnInit {
   }
 
   private isFieldDisabled(field: FilterField): boolean {
-    if (field.type === undefined || field.type === 'text') {
+    if (field.type === FilterType.TEXT) {
       return false;
     }
-    return (field.queries === undefined || field.queries.length === 0);
+    return (field.queries === undefined || field.type === undefined);
   }
 
   private queryInputChange(value: string) {
--- a/src/app/toolbar/examples/toolbar-example.component.ts
+++ b/src/app/toolbar/examples/toolbar-example.component.ts
@@ -35,6 +35,7 @@ export class ToolbarExampleComponent implements OnInit {
   currentSortField: SortField;
   toolbarConfig: ToolbarConfig;
   weekDayQueries: any[];
+  continentQueries: any[];
 
   monthVals: any = {
     'January': 1,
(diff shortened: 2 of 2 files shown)


src/app/filter/filter-fields.component.ts
```diff
@@ -219,10 +220,10 @@ export class FilterFieldsComponent implements DoCheck, OnInit {
   }
 
   private isFieldDisabled(field: FilterField): boolean {
-    if (field.type === undefined || field.type === 'text') {
+    if (field.type === FilterType.TEXT) {
       return false;
     }
-    return (field.queries === undefined || field.queries.length === 0);
+    return (field.queries === undefined || field.type === undefined);
```
dlabrecq, line 226: It's not clear from the description, but I suspect you're making these changes to generate an onFieldSelect event? I'm not sure we want to introduce this change here.

My first concern is that the field is no longer disabled when there are no queries. This may address your issue, but suspect this change will impact other users of patternfly-ng? I'm fairly certain this will effect the save filter feature.

In addition, the field.type === undefined check has been moved. Now when devs don't specify field.type, this function will instead return true instead of false, which is not what we want.

That said, we recently introduced a reset() method to FilterComponent. This method is public, but not currently exposed via the toolbar -- I'll add that next.

FYI, we already have an example of this for the save filter function. This same scenario occurs here when there are no more saved filters (queries) available. Thus, the selected filter field must be reset.

https://github.com/patternfly/patternfly-ng/blob/a1641489802eef7e36fd5f257c97874ae1585508/src/app/filter/examples/filter-save-example.component.ts#L351

dlabrecq (changes requested):

src/app/filter/filter-fields.component.ts
dlabrecq: Filter reset is now available via toolbar. Please see: https://github.com/patternfly/patternfly-ng/pull/258

vikram-raj: > My first concern is that the field is no longer disabled when there are no queries

Yes, If devs pass queries of length zero then the field should not be disabled. Because it will give an option to dev to build the queries when the user selects the field. For example, in the planner, we just load the field on ngOnInIt with queries of length zero and build the queries after selecting the field.

> I'm fairly certain this will effect the save filter feature.

I do check the save filter feature and it is working as expected.

> In addition, the field.type === undefined check has been moved. Now when devs don't specify field.type, this function will instead return true instead of false, which is not what we want.

I think field.type is mandatory, devs have to specify the field.type otherwise how can component knows that its typeahead, select or text field type.

> FYI, we already have an example of this for the save filter function. This same scenario occurs here when there are no more saved filters (queries) available. Thus, the selected filter field must be reset.

No, we have a different scenario i:e we don't want to disable the field if there is no queries available because we make queries when user select the field.

sanbornsen: Yeah basically, this toolbar should bar should be able to support lazy loading of queries.

dlabrecq: The field type is an optional property.

Unfortunately, when I add 'queries: []' to the save filter example, the "My Filters" option is no longer disabled as expected. The intention is that the filter field option should be disabled when there are no queries.

FYI, the save filter example also loads queries lazily. However, the filter field selection isn't used to fetch those queries. There is also a "Lazy Loading" example showing lazily loaded filter fields. 

That said, if you want to enable the filter field option, you can create a dummy query object. The value doesn't matter because you're example immediately replaces the queries anyway. For example, you could add this to your config:

queries: [{value: 'placeholder'}]

Alternatively, a disabled=false property could be created to override the default behavior. However, the default behavior must be the same for existing use cases.

sanbornsen: That sounds like a good idea.

vikram-raj: > queries: [{value: 'placeholder'}]

Yes, It's a good Idea. Will do this. Hence closing this PR.

PR #276, opened 2018-01-15
fix(sparkline): avoid regenerating chart on every data update

### LINKED ISSUE #275: Sparkline charts "flicker" on data update
When changing the data set of a Sparkline chart component, the chart has a visual flicker or apparent redraw. This is visible on the demo page here:

http://www.patternfly.org/patternfly-ng/#/sparkline

Clicking the "Add Data Point" button adds data to the chart and presents the flicker/redraw behaviour, and the chart data does not animate a smooth update as demonstrated by c3 here:

http://c3js.org/samples/data_load.html

andrewazores:
When only chart data has changed and config remains unchanged, the sparkline chart component will
not be regenerated but the data simply reloaded instead. This should be a lighter-weight operation
and avoids a visual flicker and redraw of the chart component on data update.

fixes https://github.com/patternfly/patternfly-ng/issues/275

### DIFF
--- a/src/app/chart/sparkline/sparkline.component.ts
+++ b/src/app/chart/sparkline/sparkline.component.ts
@@ -60,9 +60,10 @@ export class SparklineComponent extends ChartBase implements DoCheck, OnInit {
    * Check if the component config has changed
    */
   ngDoCheck(): void {
-    if (!isEqual(this.config, this.prevConfig) || !isEqual(this.chartData, this.prevChartData)) {
+    const dataChanged = !isEqual(this.chartData, this.prevChartData);
+    if (dataChanged || !isEqual(this.config, this.prevConfig)) {
       this.setupConfig();
-      this.generateChart(this.config, true);
+      this.generateChart(this.config, !dataChanged);
     }
   }
 


andrewazores:
The existing demo page for the Sparkline component acts as a good manual reproducer for this issue. With this patch applied, clicking "Add Data Point" results in the chart animating a data point addition rather than entirely redrawing.

src/app/chart/sparkline/sparkline.component.ts
```diff
@@ -61,8 +61,9 @@ export class SparklineComponent extends ChartBase implements DoCheck, OnInit {
    */
   ngDoCheck(): void {
     if (!isEqual(this.config, this.prevConfig) || !isEqual(this.chartData, this.prevChartData)) {
+      const dataChanged = isEqual(this.chartData, this.prevChartData);
```
dlabrecq: Can this line be placed outside the if-statement? That way, isEqual(this.chartData, this.prevChartData)) is only called once.

dlabrecq (changes requested):

src/app/chart/sparkline/sparkline.component.ts
andrewazores: Done.

dlabrecq (approved):
LGTM

PR #292, opened 2018-01-19
chore(sort-tests): remove unnecessary statements in unit test

seanforyou23:
for sort component

when relying on tick to simulate passage of time, we only need
to call fixture.detectChanges() after the tick() statement

tested on ngx-bootstrap v1.8.0 and 1.9.3

If this is kosher, will post another PR for resolving this in the remaining tests

### DIFF
--- a/src/app/sort/sort.component.spec.ts
+++ b/src/app/sort/sort.component.spec.ts
@@ -62,9 +62,8 @@ describe('Sort component - ', () => {
 
     let button = element.querySelector('.sort-pf button');
     button.click();
-    fixture.detectChanges(); // Workaround to fix dropdown tests
     tick();
-    fixture.detectChanges();
+    fixture.detectChanges(); // Workaround to fix dropdown tests
 
     let elements = element.querySelectorAll('.sort-pf .sort-field');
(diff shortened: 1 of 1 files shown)


src/app/sort/sort.component.spec.ts
```diff
@@ -62,7 +62,6 @@ describe('Sort component - ', () => {
 
     let button = element.querySelector('.sort-pf button');
     button.click();
-    fixture.detectChanges(); // Workaround to fix dropdown tests
     tick();
```
dlabrecq, line 65: Can you please add the "Workaround to fix dropdown tests" comment. When folks cut/paste tests, I want to be clear as to why this is being used.

dlabrecq (dismissed):
LGTM

src/app/sort/sort.component.spec.ts
seanforyou23: No problem, I didn't know if we were still considering it a workaround :) I'll add this comment back.

dlabrecq (approved):
LGTM

dlabrecq (approved):
LGTM

PR #296, opened 2018-01-19
chore(filter-tests): remove unnecessary statements in unit test

seanforyou23:
when relying on tick to simulate passage of time, we only need
to call fixture.detectChanges() after the tick() statement

tested with ngx-bootstrap v1.8.0 and 1.9.3


dlabrecq (approved):
LGTM

PR #302, opened 2018-01-20
Bump ngx-bootstrap version

dlabrecq:
Bumped ngx-bootstrap to v2.0.0 and made it an optional dependency.

Not all patternfly-ng modules require ngx-bootstrap (e.g., chart, empty state, about modal, remaining-chars, and pipes). Packages such as ngx-datatable, angular-tree-component, c3, and ng2-dragula are already defined as optional dependencies.

Tested with OSIO, which still uses ngx-bootstrap v1.8.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,14 +67,14 @@
   },
   "dependencies": {
     "lodash": "^4.17.4",
-    "ngx-bootstrap": "^1.8.0",
     "patternfly": "^3.28.0"
   },
   "optionalDependencies": {
     "@swimlane/ngx-datatable": "^11.1.7",
     "angular-tree-component": "^6.0.0",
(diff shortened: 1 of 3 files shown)


seanforyou23:
Just a quick update: I've done some basic testing and everything generally look great with this PR. 

Another important test would be to run this against Syndesis UI, as it was one of the driving factors for making this update (among others), however I haven't been able to verify this yet as we're in the middle of upgrading quite a few related dependencies all at once.

I'm cool with this being merged if its needed for anything else, otherwise holding off long enough to finish the Syndesis UI upgrades might be good as we can then give the stamp of approval with more confidence than I have at the moment.

seanforyou23 (approved):
Looks good to me!

PR #303, opened 2018-01-21
Patternfly as a peer dependency

dlabrecq:
Made Patternfly a peer dependency because patterfnly-ng never references those files directly -- except during build time. 

Similar to Angular and Typescript, there can only be one version ever used during run time. Additionally, Patternfly CSS and patternfly-settings.js must be built and/or included manually, by the app.

Tested with OSIO, which builds its own patternfly v3.30.1.


dlabrecq:
Talked to @dgutride and we decided to leave this one as is for now.

PR #304, opened 2018-01-21
Each module should have its own index.ts

dlabrecq:
I would like to be able to import just the notifications module; for example, without my app also having to build all of patternfly-ng. That is, I want the ability to import any given module without installing unused packages like ngx-datatable, angular-tree-component, c3, and ng2-dragula. 

As it is, patternfly-ng cannot be built unless all packages are installed in my app. However, when each module has its own index.ts, packages such as @swimlane/ngx-datatable can truly be optional.

Currently, all of patternfly-ng is included in the app build when the a module is specified like this:

`import { NotificationModule } from 'patternfly-ng';`

The postinstall node script copies modules to the root directory, allowing individual patternfly-ng modules to be imported like so:

`import { NotificationModule } from 'patternfly-ng/notification';`

Tested with `ng build` and `ng build --prod`


dlabrecq:
This has been merged with #306

PR #306, opened 2018-01-23
Remove embedded CSS

dlabrecq:
This is a breaking change that removes embedded CSS from the components. This allows LESS to be customized, but requires the app to now include patternfly-ng.min.css. 

HTML templates are still embedded in the component.

FYI, changes are based on https://github.com/patternfly/patternfly-ng/pull/304, so the PR shows those changes as well -- that PR moves modules to the package root.

Note that the dist directory structure more closely resembles angular-patternfly.
- There's a 'less' folder containing all the patternfly-ng files necessary to build your own CSS.
- There's a 'css' folder, which includes patternfly-ng.css and patternfly-ng.min.css


AdamJ (approved):
LGTM - I ran the build locally and the demo worked without issue.

Issue #146, opened 2017-09-30
Notification drawer pattern


catrobson:
Convert notification drawer from angular-patternfly
http://www.patternfly.org/angular-patternfly/#/api/patternfly.notification.component:pfNotificationDrawer

http://www.patternfly.org/pattern-library/communication/notification-drawer/

akoserwal:
@catrobson: I would like to work on this pattern.

catrobson:
@akoserwal sounds great!  Do you need anything to get started?

akoserwal:
@catrobson: I can begin now, got some understanding with application launcher work.

akoserwal:
@catrobson : we can close this issue.

dlabrecq:
PR https://github.com/patternfly/patternfly-ng/pull/334 has been merged

PR #321, opened 2018-02-14
WIP: Feature pf notification drawer issue: #146

akoserwal:
#146  
Few things need to clarify before moving forward, for these parameters, do we want to take template input from user ?
ng-include: directive was used in these params
- headingInclude
- subheadingInclude 
- notificationBodyInclude 
- notificationFooterInclude
https://www.patternfly.org/angular-patternfly/#/api/patternfly.notification.component:pfNotificationDrawer

### DIFF
--- a/src/app/notification/index.ts
+++ b/src/app/notification/index.ts
@@ -1,6 +1,8 @@
 export { InlineNotificationComponent } from './inline-notification/inline-notification.component';
 export { Notification } from './notification';
+export { NotificationDrawerComponent  } from './notification-drawer/notification-drawer.component';
 export { NotificationEvent } from './notification-event';
+export { NotificaitonGroup } from './notification-group';
 export { NotificationModule } from './notification.module';
 export { NotificationType } from './notification-type';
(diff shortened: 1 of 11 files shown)


dlabrecq:
Yes, please use ng-template. 

Please rename headingInclude, titleInclude, etc. as headingTemplate and titleTemplate. That would be consistent with the naming convention used in other components.

We can also drop the "drawer" prefix for most properties -- the component is already named "notification-drawer". For example, drawerTitle, drawerExpanded, and drawerHidden can be simplified as title, expanded, visible/show.

Also, please avoid using function call backs (i.e., actionButtonCallback) -- patternfly-ng uses events instead. For example, the toolbar, list-base, and the action component all provide an onActionSelect event for buttons. Alternatively, another template could be used here to provide your own button?

AdamJ:
@akoserwal Please close this PR until you are ready for a full review. If you'd like us to look at something that is in progress, you can file a PR against your own forked master branch, and we can look there. Thanks!

akoserwal:
@mindreeper2420: Sure

PR #323, opened 2018-02-28
Expose ngx-datatable events; detailToggle, page, and sort

dlabrecq:
Expose ngx-datatable events; detailToggle, page, and sort. This will allow developers to more easily bypass the default implementation.

Note that these particular events are not compatible with the default pagination and sorting provided by pfng-table. Thus, I've updated the typescript docs to reflect that.

Also locked the typescript dev dependency at v2.6.2 in order for the Travis build to succeed. Using v2.7 breaks lodash.


elvisisking (approved):
Looks good @dlabrecq !

dgutride (approved):

Issue #342, opened 2018-05-03
Create wrapper component for @ng-select/ng-select


ghost:
I want to use a smart select component, and `@ng-select/ng-select` fits the bill, but doesn't match PatternFly's style.

dlabrecq:
FYI, the patternfly-ng team discussed the @ng-select/ng-select stylesheet you provided via https://github.com/patternfly/patternfly-ng/pull/343 -- thank you for contributing that.

Unfortunately, we don't want to support stylesheets for external libraries at this time. There are a few pros/cons with this unique approach that's worth exploring further. However, forcing patternfly-ng users go to another library to fetch the component didn't sit well with the team.

It's possible that we wrap @ng-select/ng-select as a component in patternfly-ng. However, we need to take a closer look at that library to see if it's a good fit for PatternFly.

dlabrecq:
Closing due to inactivity.

PR #343, opened 2018-05-03
Add a stylesheet for @ng-select/ng-select

ghost:
Here are the styles I've created for `@ng-select/ng-select`. Please note that I'm not a designer, nor am I particularly familiar with PatternFly, so there's likely some stuff that I've done wrong. Let me know and I'll fix it all.

See #342


dlabrecq:
Added some notes in https://github.com/patternfly/patternfly-ng/issues/342 to track our discussions.

dlabrecq:
Closing since we will not be supporting stylesheets for external libraries. However, we'll keep https://github.com/patternfly/patternfly-ng/issues/342 open to retain a copy of these styles.

PR #351, opened 2018-05-17
feat(clipboard-copy): add inline/block level copy component, copy service

seanforyou23:
This PR adds two new components, and a service. The two components are an inline and block level variation of a "copy value to clipboard" functionality. These components are simple, flexible, and accessible. They are also WIP! The basic functionality is there, feel free to test drive and make suggestions!

screen shot 2018-05-16 at 10 53 05 pm

https://rawgit.com/seanforyou23/patternfly-ng/copy-clipboard-dist/dist-demo/
https://rawgit.com/seanforyou23/patternfly-ng/copy-clipboard-dist/dist-demo/#/inlinecopy
https://rawgit.com/seanforyou23/patternfly-ng/copy-clipboard-dist/dist-demo/#/blockcopy
https://rawgit.com/seanforyou23/patternfly-ng/copy-clipboard-dist/dist-demo/#/copyservice

create block-level copy component
create inline copy component
create copy module
create copy service
add new components and service to demo navigation
accessibility considerations
add unit tests for new components and service
copy service examples
label is only required input prop
support multiple copy components per view
ensure semantic relation between label and copy button
use button instead of repurposing icon element
add tooltips for hover state
examples that use callbacks to launch toast notifications
tslint rules
expanded
dedicated tooltip and aria label inputs
clean up examples
export * as modules
align markup between inline and block versions
basic, expanded, wrap, multiple, and callback examples


src/app/copy/block-copy/block-copy.component.less
```diff
@@ -0,0 +1,55 @@
+.copy {
```
dlabrecq: We should prefix patternfly-ng selectors with `pfng-`, unless this is extending an existing patternfly selector?

src/app/copy/inline-copy/inline-copy.component.less
```diff
@@ -0,0 +1,31 @@
+.inline-copy {
```
dlabrecq: We should prefix this with `pfng-`

src/app/copy/inline-copy/inline-copy.component.ts
```diff
@@ -0,0 +1,47 @@
...
+
+@Component({
+  encapsulation: ViewEncapsulation.None,
+  selector: 'pfng-inline-copy',
+  templateUrl: './inline-copy.component.html',
+  styleUrls: ['./inline-copy.component.less']
+})
+
+export class InlineCopyComponent implements OnInit {
+  @Input('label') label: string;
```
dlabrecq: The label name is misleading since it appears to be used for the aria-label. I thought this was for the button label until I looked at the HTML template. 

Can we rename this as `ariaLabel`? We may want to add a unique `tooltip` input as well, considering the tooltip and aria-label are different.

src/app/copy/inline-copy/inline-copy.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  @Input('token') token: string = 'Missing \'token\' @Input property';
```
dlabrecq: I imagine users will want to copy more than tokens? Can we rename this as `value` or perhaps `copyValue`?

src/app/copy/inline-copy/inline-copy.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  @Input('copyBtnTxt') copyBtnTxt: string = 'Copy';
```
dlabrecq: Can we rename this as `buttonLabel`? That would be more consistent with other components.

src/app/copy/inline-copy/example/inline-copy-example.component.html
```diff
@@ -0,0 +1,50 @@
+<div class="padding-15">
```
dlabrecq, line 1: Can we split the examples under individual tabs? Similar to basic cards, donut chart, etc.

These particular examples are not complicated, but the page looks a little crowded. It would also help clarify things a bit since each would have its own js and typescript files.

src/app/copy/inline-copy/inline-copy.component.html
```diff
@@ -0,0 +1,16 @@
...
+  <span
+    class="token-cont"
+    data-toggle="tooltip"
+    data-placement="top"
+    [tooltip]="label">
+    {{token}}
+  </span>
+  <button
+    class="copy-btn"
+    [attr.aria-label]="copyBtnTxt + ' ' + label"
```
dlabrecq: I don't know that we should be concatenating strings for the aria-label, considering this is a user input -- especially if label is undefined.

I believe we should have a separate `ariaLabel` input and use it as is.

src/app/copy/copy.module.ts
```diff
@@ -0,0 +1,24 @@
...
+@NgModule({
+  imports: [
+    CommonModule,
+    TooltipModule.forRoot()
+  ],
+  declarations: [BlockCopyComponent, InlineCopyComponent],
+  exports: [BlockCopyComponent, InlineCopyComponent],
+  providers: [CopyService]
+})
+export class CopyModule {}
```
dlabrecq: We need to start providing modules per component so they can be imported individually. Can you please remove this module and add one for each component? Each component should also have its own index.js file.

src/app/copy/index.ts
```diff
@@ -0,0 +1,5 @@
+export { CopyModule } from './copy.module';
+export { CopyService } from './copy-service/copy.service';
+export { BlockCopyComponent } from './block-copy/block-copy.component';
+export { InlineCopyComponent } from './inline-copy/inline-copy.component';
+
```
dlabrecq: I'd like each component to have its own index.js file. This file would then look more like this:

```
export { 
  CopyService,
  CopyServiceModule
} from './copy-service/index';

export { 
  BlockCopyComponent,
  BlockCopyModule
} from './block-copy/index';

export {
  InlineCopyComponent,
  InlineCopyModule
} from './inline-copy/index';
```

src/app/index.ts
```diff
@@ -37,6 +37,14 @@ export {
   SparklineData
 } from './chart/index';
 
+// CopyToClipboard
+export {
+  CopyModule,
+  CopyService,
+  BlockCopyComponent,
+  InlineCopyComponent
+} from './copy/index';
```
dlabrecq: I'd like to create modules for each component, so we'll need to add exports for them here as well.

```
export {
  BlockCopyComponent,
  BlockCopyModule,
  CopyService,
  CopyServiceModule,
  InlineCopyComponent,
  InlineCopyModule
} from './copy/index';
```

src/app/copy/copy-service/copy.service.ts
```diff
@@ -0,0 +1,45 @@
...
+    } finally {
+      if (textarea.parentNode !== undefined) {
+        textarea.parentNode.removeChild(textarea);
+      }
+    }
+    return result;
+  }
+
+  private handleError(error: any): void {
+    console.error(error);
```
dlabrecq: We probably shouldn't write the error to the console by default. The notification service wraps this with a verbose variable, so you at least have to turn it on:

    if (this.verbose) {
      console.log(message);
    }

Alternatively, we could just throw the error and let the application handle it?

src/app/copy/copy-service/copy.service.ts
```diff
@@ -0,0 +1,45 @@
+import { Inject, Injectable } from '@angular/core';
+import { DOCUMENT } from '@angular/platform-browser';
```
dlabrecq: My IDE suggests DOCUMENT is deprecated? The docs say "import from `@angular/common` instead."

src/app/copy/inline-copy/inline-copy.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  @Output('copiedToClipboard') copiedToClipboard = new EventEmitter();
+
+  /**
+   * The default constructor
+   */
+  constructor(
+    private copyService: CopyService) {}
+
+  ngOnInit(): void {
+    if (!this.label) throw new Error('Missing required @Input property \'label\'');
```
dlabrecq: I don't believe we should force a tooltip or aria-label.

src/app/copy/inline-copy/inline-copy.component.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    fixture.detectChanges();
+    const ariaLabel = fixture.debugElement.children[0].children[1].attributes['aria-label'];
+    expect(ariaLabel).toBe(`${btnTxt} ${inlineCopy.label}`);
+  });
+
+  it('should throw error without a label', () => {
+    (<any>Object).assign({label: null}, inlineCopy, componentConfig);
+    expect(() => {
+      fixture.detectChanges();
+    }).toThrow(new Error('Missing required @Input property \'label\''));
```
dlabrecq: I don't believe we should force a tooltip or aria-label.

src/app/copy/inline-copy/inline-copy.component.less
```diff
@@ -0,0 +1,31 @@
...
+  display: inline-flex;
+  border: 1px solid #bbb;
+  background-color: #fff;
+  font-size: smaller;
+  margin-left: .25em;
+  margin-right: .25em;
+  max-width: 100%;
+
+  .token-cont {
```
dlabrecq: Thinking we should chose a more generic selector name? I imagine this component will be used to copy other values besides a token.

In the block-copy css, it looks like we're using `txt-cont`?

src/app/copy/block-copy/example/block-copy-example.component.ts
```diff
@@ -0,0 +1,55 @@
+import {
```
dlabrecq, line 1: I'd like to split the examples into separate tabs -- see my comment for inline-copy-example

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+   * Define format for copy button's aria label
+   */
+  copyBtnAriaLabel(componentLabel: string): string {
+    return `Copy ${componentLabel}`;
+  }
+
+  ngOnInit() {
+    if (!this.label) throw new Error('Missing required @Input property \'label\'');
```
dlabrecq: I don't believe we should force a tooltip or aria-label.

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+@Component({
+  encapsulation: ViewEncapsulation.None,
+  selector: 'pfng-block-copy',
+  templateUrl: './block-copy.component.html',
+  styleUrls: ['./block-copy.component.less']
+})
+
+export class BlockCopyComponent implements OnInit {
+  @Input('label') label: string;
```
dlabrecq: I'd like to rename this as `ariaLabel` and add a unique `tooltip` input. See my comments for inline-copy.

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,73 @@
...
+  @Input('token') token: string = 'Missing \'token\' @Input property';
```
dlabrecq: I'd like to rename this as `value` or possibly `copyValue`. See my comments for inline-copy.

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,73 @@
...
+  @Input('copyBtnTxt') copyBtnTxt: string = 'Copy';
```
dlabrecq: I'd like to rename this as `buttonLabel`. See my comments for inline-copy.

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,73 @@
...
+  @Input('tokenPanelOpen') tokenPanelOpen: boolean = false;
```
dlabrecq: I'd like to rename this as `expanded` or possibly `panelExpanded`, This would be more consistent with other components.

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  @Output('copiedToClipboard') copiedToClipboard = new EventEmitter();
+
+  /**
+   * Used to uniquly relate label to copy button
+   */
+  public hash: number = Math.floor(Math.random() * 10000);
```
dlabrecq: Does this need to be public? If so, consider renaming this as _hash and provide an accessor like:

```
get hash(): number {
  return this._hash;
}
```

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+   * The default constructor
+   */
+  constructor(private copyService: CopyService) {}
+
+  /**
+   * Toggle copy token panel open and close
+   */
+  toggleTokenPanel(): void {
```
dlabrecq: Rename as togglePanel

src/app/copy/block-copy/block-copy.component.html
```diff
@@ -0,0 +1,33 @@
+<div>
+  <label [attr.for]="'pf-copy-token-button-' + hash">{{label}}</label>
```
dlabrecq: Can you use `pfng-` as the prefix

src/app/copy/block-copy/block-copy.component.html
```diff
@@ -0,0 +1,33 @@
...
+        [attr.id]="'pf-copy-token-button-' + hash"
+        class="btn btn-lg btn-default copy-btn"
+        [attr.aria-label]="copyBtnAriaLabel(label)"
+        (click)="copyTokenToClipboard()">{{copyBtnTxt}}</button>
+    </div>
+    <div class="copy-body" *ngIf="tokenPanelOpen">
+      <span>{{token}}</span>
+    </div>
+  </div>
+</div>
```
dlabrecq: Need a new line

src/app/copy/block-copy/block-copy.component.html
```diff
@@ -0,0 +1,33 @@
...
+
+      <div class="copy-preview-txt-cont">
+        <span
+          data-toggle="tooltip"
+          data-placement="top"
+          [tooltip]="label">{{token}}</span>
+      </div>
+
+      <button
+        [attr.id]="'pf-copy-token-button-' + hash"
```
dlabrecq: Is the hash being used to create a unique ID? If so, we should talk about standardizing how we do that for patternfly-ng components. We may want to allow the application to provide its own ID.

FYI, there is an issue to provide checkbox IDs for pfng-list. Need to make sure whatever we do here is consistent.

src/app/copy/copy-service/copy.service.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  it('should return true if the dom copy command has succeeded', () => {
+    spyOn(service.dom, 'execCommand').and.returnValue(true);
+    let result = service.copy('foobar');
+    expect(result).toBeTruthy();
+  });
+
+  it('should return false if the dom copy command has failed', () => {
+    spyOn(service.dom, 'execCommand').and.throwError('expected-test-error');
```
dlabrecq, line 35: Are we throwing an error or writing to the console? Or, both?

Looks like the service is catching that...

src/app/copy/inline-copy/inline-copy.component.less
```diff
@@ -0,0 +1,31 @@
...
+    box-shadow: none;
+    background-image: linear-gradient(to bottom, #fafafa, #ededed);
+
+    &:hover,
+    &:active,
+    &:focus {
+      background-color: #f1f1f1;
+    }
+  }
+}
```
dlabrecq: Needs a new line

dlabrecq (changes requested):
Built/ran the demo locally. Looks nice, just minor comments

dlabrecq (changes requested):
Additional comments are hidden above...

src/app/copy/inline-copy/inline-copy.component.ts
seanforyou23: Sure I see, I was using label here because it aligned nicely with the same `label` input property for the block level copy component but it's a good point about it not actually being label text, I'll update this.

Allowing for custom tooltip text sounds like a good idea, I'll just set it to the ariaLabel value by default, and users can override if they like.

src/app/copy/inline-copy/inline-copy.component.ts
seanforyou23: I was thinking of "token" in the generic sense like "node" or something. `copyValue` is nice and obvious :)

src/app/copy/block-copy/block-copy.component.html
seanforyou23: Ah, yes good call. I figured that would come up and I'm all for it. You know, though, I'd really rather not use a arbitrary hash here of at all, if possible, and instead use an actual text string that expresses some semantic meaning. For example mapping over the aria label (or something like that) and appending a dash-cased-version of that to the ID. That could be really long so maybe not that specifically, but you get the idea.

I really just couldn't think of anything I liked while I was writing that so I sort of stubbed it out with this thingie.

You know, this brings up another thing I wanted to point out, which is that because the button and label are linked, if you select the label, it's the same as clicking the button. This seems on the surface like a nice thing to have, but wanted to point it out in case this causes unintended tripping hazard/frustration like "oops I didn't mean to copy, now I'm upset".

Not so big a deal for the sighted user, but for keyboard user, we have no idea how difficult/time consuming it may be to retrieve whatever value they previously had in their clipboard. I could use a span instead of label to remove this interactivity and keep the visual appearance. So that's another option.

If the act of selecting the label causing copy to clipboard action is problematic, we could get rid of the id altogether.

The original design didn't have a label, but I felt like it needed one during implementation because with more than one on the screen, you need a way to ENSURE there is some difference between the buttons, otherwise they all show up as "Copy button" or something similar depending on reader.

Dropping the visual label and instead opting for an aria-label in it's place could be another option.

If we go for the hash based approach, let's use an incrementing counter over random string?

seanforyou23: @dongniwang or @jgiardino may know of a tie breaker. Definitely worth checking if there's a single solution for both this and pfng-list.

src/app/copy/copy-service/copy.service.spec.ts
seanforyou23: Yea I think that description is from copy pasta, good catch! I'll clean this up with the rest of the tests

jgiardino:
These examples look great! I love that you have `aria-label` on the icon buttons 👍 
```
<button class="copy-btn" aria-label="Copy Example Sparql Query">
    <i class="fa fa-copy"></i>
  </button>
```
But since the icon provides no meaning to a screen reader, you can also add `aria-hidden="true"`:
`<i class="fa fa-copy" aria-hidden="true"></i>`

src/app/copy/block-copy/block-copy.component.html
jgiardino: Can you describe the user flow you're asking about? I'm not really following what the behavior in question is.

Although while trying to figure it out, I did notice that when I have focus on a Copy button and hit Enter, the focus isn't maintained on the Copy button. I would expect that focus to remain on the action I selected unless there's a reason to shift focus.

src/app/copy/inline-copy/inline-copy.component.html
seanforyou23: Yea I was definitely noticing some limitations of this approach when I was writing it. That said, it did have the nice effect of 1) ensuring a sensible, and unique accessible name was generated for each control and 2) we know that the experience will be uniform for this type of control to the screen reader user.

I see though why it's good not to make assumptions about how users of the framework will want to express their content to their readers. Maybe we could make this a setting, that's set by default to ensure a baseline level of accessibility is provided by the framework, and then the ability to turn it off and use your own aria attributes if you want?

For now i'll focus on using ariaLabel as an input as is and we can think about it from there.

seanforyou23: Now that I think about it might be good to just show this as an example or something. I'll come up with something more fleshed out and we can go from there

src/app/copy/block-copy/block-copy.component.html
seanforyou23: Sure, so the flow would be; a user clicks on the text label for the block copy component, which causes the text to be copied to their clipboard.

In general, for form controls, it's expected that when a user clicks the label, it will result in the same behavior as if the user clicks the corresponding form control directly. The more common scenario is when a user clicks a label resulting in focus placed in a related input. With the related control being a button, though, the result is whatever happens when you click the button.

Would this be considered unexpected behavior? I'm starting to think it may be, but wanted another opinion. It's easy to make this work either way, just wondering what's best.

Noticed the focus leaving the button on keypress, I'll fix this. Good catch.

src/app/copy/inline-copy/inline-copy.component.ts
dlabrecq: I don't believe either should be required. If a tooltip is not provided, lets not output one.

src/app/copy/block-copy/block-copy.component.html
dlabrecq: My concern is more about how do we generate IDs for patternfly-ng overall. I'm thinking if the user provides an ID, we should use that as a prefix. Lodash has a uniueid method we could use by default if an ID isn't given.

src/app/copy/inline-copy/inline-copy.component.ts
seanforyou23: I see, in an effort to keep the component as flexible as possible, I'll change this so it doesn't throw an error.

One benefit of making this required was that I was then able to construct sensible defaults for the aria labels, in case users didn't add them theirselves. The problem that came up was when there were several being used on a same page, from an accessibility perspective you couldn't tell them apart. "Copy Button, Copy Button, Copy Button" for example. So, it was kinda nice being able to ensure this aspect of accessibility wasn't neglected.

But yea, I'll remove this check from the core component and just try to hit that point home with one of the examples.

seanforyou23:
Thanks for the awesome feedback everyone, I've addressed a good bit of it but still have some cleaning up to do, particularly around the examples and tests - but definitely a little more component work as well. More feedback is welcome while I'm working through those things though!

src/app/copy/block-copy/block-copy.module.ts
```diff
@@ -0,0 +1,24 @@
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+
+import { TooltipConfig, TooltipModule } from 'ngx-bootstrap/tooltip';
+
+import { CopyService } from '../copy-service/copy.service';
+import { BlockCopyComponent } from './block-copy.component';
+
+export {
+  BlockCopyComponent
```
dlabrecq: Don't need to export the component because it's already export below.

src/app/copy/block-copy/block-copy.module.ts
```diff
@@ -0,0 +1,24 @@
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+
+import { TooltipConfig, TooltipModule } from 'ngx-bootstrap/tooltip';
```
dlabrecq, line 4: TooltipConfig is unused. We need to add TooltipConfig as a provider, unless we don't really need a tooltip here?

If you do that here, it may not be required in the example? If that can be removed from the examples, all the better.

src/app/copy/block-copy/block-copy.component.html
```diff
@@ -0,0 +1,36 @@
+<div class="pfng-block-copy">
+  <label
+    class="pfng-block-copy-label"
+    [attr.for]="'pfng-block-copy-value-button-' + hash">{{label}}</label>
```
dlabrecq: If the label has not been provided, I'd like to make this optional (i.e., add an *ngIf statement here).

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,68 @@
...
+      let result = this.copyService.copy(this.copyValue);
+      if (result) {
+        this.copiedToClipboard.emit(`${this.label} copied!`);
+      } else {
+        console.error(`Failed to copy ${this.label}`);
+      }
+  }
+
+  ngOnInit() {
+    if (!this.label) throw new Error('Missing required @Input property \'label\'');
```
dlabrecq: The label isn't something that should be required.

Let's not throw an error here, but provide defaults for copyBtnAriaLabel and expandBtnAriaLabel instead. Alternatively, we can use the button label, if provided?

The button label is most likely to always be used, but neither should throw an error.

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  @Input('copyValue') copyValue: string = 'Missing \'copyValue\' @Input property';
+  @Input('buttonLabel') buttonLabel: string = 'Copy';
+  @Input('expanded') expanded: boolean = false;
+
+  @Output('copiedToClipboard') copiedToClipboard = new EventEmitter();
+
+  /**
+   * Used to uniquly relate label to copy button
+   */
+  public _hash: number = Math.floor(Math.random() * 10000);
```
dlabrecq: Can we use the uniqueid method from lodash?

src/app/copy/copy-service/copy.service.ts
dlabrecq: The deprecation is still an issue. Did you look at importing DOCUMENT with @angular/common? Seems to work for me.

src/app/copy/inline-copy/inline-copy.module.ts
```diff
@@ -0,0 +1,24 @@
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+
+import { TooltipConfig, TooltipModule } from 'ngx-bootstrap/tooltip';
```
dlabrecq, line 4: TooltipConfig is unused. Need to add this to the provider. 

Need to check, but we may be able to remove this from the example modules if it's provided here?

src/app/copy/inline-copy/inline-copy.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  encapsulation: ViewEncapsulation.None,
+  selector: 'pfng-inline-copy',
+  templateUrl: './inline-copy.component.html',
+  styleUrls: ['./inline-copy.component.less']
+})
+
+export class InlineCopyComponent implements OnInit {
+  @Input('ariaLabel') ariaLabel: string;
+  @Input('copyValue') copyValue: string = 'Missing \'copyValue\' @Input property';
+  @Input('tooltipTxt') tooltipTxt: string;
```
dlabrecq: Please rename this to tooltip.

We may also want to provide a tooltipPlacement input so devs can specify how it's shown. We don't always have room at the top.

src/app/copy/inline-copy/inline-copy.component.html
```diff
@@ -0,0 +1,16 @@
+<span
+  class="pfng-inline-copy">
+  <span
+    class="pfng-inline-copy-txt-cont"
+    data-toggle="tooltip"
+    data-placement="top"
```
dlabrecq: data-toggle and data-placement are old attributes. Please use the ngx-bootstrap attributes; placement and tooltip.

src/app/copy/inline-copy/inline-copy.component.html
```diff
@@ -0,0 +1,16 @@
...
+    [attr.tooltip]="tooltipTxt">
```
dlabrecq: If you're using the HTML tooltip attribute, we don't need the ngx-bootstrap functionality. However, I'm thinking we want to use ngx-bootstrap for consistency.

src/app/copy/inline-copy/inline-copy.component.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * The default constructor
+   */
+  constructor(private copyService: CopyService) {}
+
+  ngOnInit(): void {
+    if (!this.ariaLabel) throw new Error('Missing required @Input property \'ariaLabel\'');
+    if (!this.tooltipTxt) {
+      this.tooltipTxt = this.ariaLabel;
```
dlabrecq: Please do not throw an error here for ariaLabel. The ariaLabel should be optional or we can set a default.

We should also not set a tooltip by default.

src/app/copy/block-copy/block-copy.component.html
```diff
@@ -0,0 +1,36 @@
...
+          class="fa pfng-block-copy-preview-icon"
+          [ngClass]="{'fa-angle-down': expanded, 'fa-angle-right': !expanded}"></i>
+      </button>
+
+      <div class="pfng-block-copy-preview-txt-cont">
+        <span
+          class="pfng-block-copy-preview-txt"
+          data-toggle="tooltip"
+          data-placement="top"
+          [attr.tooltip]="tooltip">{{copyValue}}</span>
```
dlabrecq: We're using both ngx-bootstrap and HTML tooltip attributes -- we only need one or the other.

data-toogle and data-placement should be replaced with ngx-bootstrap attributes; placement and tooltip.

See: https://ngx-bootstrap.surge.sh/#/tooltip

dlabrecq (changes requested):
Looking good. I know this is still a WIP, but here is some initial feedback. 

I also found a small CSS issue -- missing border when expanded.

screen shot 2018-05-21 at 11 57 00 am

src/app/copy/copy-service/copy.service.ts
seanforyou23: I hadn't yet addressed this in that last commit I pushed, was getting all the familiar things out of the way first :)

I have it in place now and seems to be working just fine! Will be included in this next commit.

src/app/copy/block-copy/block-copy.module.ts
seanforyou23: Good call, I've been having some issues with tooltips and this may be part of it. Checking into it now, thx!

src/app/copy/block-copy/block-copy.component.ts
seanforyou23: Good suggestions - I'm with ya. It's sort of where I was headed initially when I was requiring the label, and using that to construct unique aria label defaults. For screen readers, it's important for the aria labels to be unique per component because if they are not, and there are more than one in a single page, it's impossible to tell the copy/expand buttons apart from one another.

I'm starting to think it's a problem we shouldn't try to solve for people at the component level, and instead just leave the component in a most flexible state. It doesn't ensure the component will be "accessible" in all scenarios, but it may be a reservation we have to make.

I'll keep thinking about this, in the mean time I'll go ahead and push my next set of updates.

src/app/copy/inline-copy/inline-copy.component.html
seanforyou23: I didn't realize those where old, good catch! I was a little hesitant to use placement/tooltip at first because they are not valid html5 attributes and are reported as error in tools like markup validation service - https://validator.w3.org/, but to be fair it's really just a conformance rule that everyone breaks anyway. I'll go with the ngx-bootstrap attributes for now.

src/app/copy/block-copy/block-copy.component.html
seanforyou23: If we were to use a provided "id" as a prefix, then it becomes, well, not the id anymore. I'm wondering what value this would add from a users perspective. It shouldn't really matter since it's a generated id anyway, right? I could be missing something..

I'm starting to think uniqueId primed with "pfng-block-copy-button-" does the job well enough and is inline with the the work you did over in list recently - https://github.com/patternfly/patternfly-ng/pull/360/

One caveat - I appended an extra dash to the prefix in my latest changes because I think it looks nice with the counter being separated with a dash like the rest of the string, but I could be convinced otherwise :)

@dlabrecq let me know what you think, happy to provide a user defined id prefix, just really wanna understand what this might be used for before I put it in there. Thx!

seanforyou23:
Awesome feedback everyone. I've just pushed another round of updates, I know there's some cleanup to do regarding tests, but it's ready for another round of review whenever anyone finds the time! Thx in advance.

src/app/copy/block-copy/block-copy.component.html
dlabrecq: Being able to provide an ID allows users to write tests based on something that is known to them. For example, if someone wrote a Selimium test to activate the button, the test would need to rely on a consistent ID. A unique, generated ID isn't something a test can rely on because it is always different. Using a generated ID is fine by default, but we need to allow for that to be overridden. The convention used by the components should be consistent -- please follow #360.

src/app/copy/inline-copy/inline-copy.component.less
```diff
@@ -0,0 +1,29 @@
...
+  border: 1px solid #bbb;
+  background-color: #fff;
+  font-size: smaller;
+  margin-left: .25em;
+  margin-right: .25em;
+  max-width: 100%;
+  &-txt-cont {
+    flex-grow: 1;
+    display: inline;
+    align-items: center;
```
mcoker: `display: inline;` looks unnecessary `.pfng-inline-copy-txt-cont` and `align-items: center;` doesn't do anything since it isn't a flex parent.

src/app/copy/block-copy/block-copy.component.less
```diff
@@ -0,0 +1,50 @@
+.pfng-block-copy {
+  &-inner-container {
+    border: 1px solid #bbb;
+  }
+  &-preview {
+    display: flex;
+    align-items: stretch;
```
mcoker: `align-items: stretch;` is the default value for `align-items` so that can be removed.

mcoker:
This looks great @seanforyou23!! 

One bit of feedback about the [block level copy example](https://rawgit.com/seanforyou23/patternfly-ng/copy-clipboard-dist/dist-demo/#/blockcopy) - have you considered adding a right border/divider between the arrow on the left and the text to copy? Currently it kinda looks like the arrow is part of the text that will be copied.

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  styleUrls: ['./block-copy.component.less']
+})
+
+export class BlockCopyComponent extends CopyBase {
+  @Input('label') label: string;
+  @Input('expandBtnAriaLabel') expandBtnAriaLabel: string;
+  @Input('buttonLabel') buttonLabel: string = 'Copy';
+  @Input('expanded') expanded: boolean = false;
+
+  public uniqueID: string = uniqueId('pfng-block-copy-button-');
```
dlabrecq: I'd like to make this equal to `uniqueId('pfng-block-copy');`. That way, lodash's unique ID is not appended to the end of the string.

See below...

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  constructor(protected copyService: CopyService) {
+    super(copyService);
+  }
+
+  /**
+   * Used to uniquly relate label to copy button
+   */
+  get copyBtnId(): string {
+    return this.uniqueID;
```
dlabrecq: Then, this can return `this.uniqueID + '-button';`

The resulting ID would look more like `pfng-block-copy1-button`. And anywhere else we need an ID later, we could create something like `pfng-block-copy1-arialabel`, for example. 

This makes testing easier if each ID is prefixed similarly. This would also follow the convention in pfng-list.

src/app/copy/block-copy/block-copy-config.ts
```diff
@@ -0,0 +1,27 @@
+import { CopyConfigBase } from '../copy-config-base';
```
dlabrecq: Considering the components are using inputs, this appears to be unused? 

I'd like to remove the config classes and move the typedoc comments to each component input.

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  selector: 'pfng-block-copy',
+  templateUrl: './block-copy.component.html',
+  styleUrls: ['./block-copy.component.less']
+})
+
+export class BlockCopyComponent extends CopyBase {
+  @Input('label') label: string;
+  @Input('expandBtnAriaLabel') expandBtnAriaLabel: string;
+  @Input('buttonLabel') buttonLabel: string = 'Copy';
+  @Input('expanded') expanded: boolean = false;
```
dlabrecq: Please add doc comments for each input.

src/app/copy/block-copy/example/block-copy-callback-example.component.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Notification } from '../../../notification/notification';
+import { NotificationService } from '../../../notification/notification-service/notification.service';
+import { NotificationType } from '../../../notification/notification-type';
+
+@Component({
+  encapsulation: ViewEncapsulation.None,
+  selector: 'block-copy-callback-example',
+  templateUrl: './block-copy-callback-example.component.html'
+})
+export class BlockCopyCallbackExampleComponent implements OnInit {
```
dlabrecq: Can we rename this example as "notification" Vs "callback"?

It was a little confusing to see the word "callback" in the example. Callbacks were something heavily used in Angular Patternfly. Instead of invoking user provided functions, we've been using events in patternfly-ng.

Considering the example is showing a notification, that may be more appropriate?

src/app/copy/copy-config-base.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * A config containing properties for copy components
+ */
+export class CopyConfigBase {
```
dlabrecq: I believe we can remove this config class and move the typedoc comments to CopyBase?

src/app/copy/inline-copy/inline-copy-config.ts
```diff
@@ -0,0 +1,6 @@
+import { CopyConfigBase } from '../copy-config-base';
+
+/**
+ * A config containing properties for inline copy
+ */
+export class InlineCopyConfig extends CopyConfigBase {}
```
dlabrecq: I believe this can be removed

src/app/copy/inline-copy/inline-copy.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  styleUrls: ['./inline-copy.component.less']
+})
+
+export class InlineCopyComponent extends CopyBase {
+
+  constructor(protected copyService: CopyService) {
+    super(copyService);
+  }
+
+  copyValueToClipboard(): void {
```
dlabrecq: Can this method be marked protected?

src/app/copy/copy-base.ts
```diff
@@ -0,0 +1,59 @@
...
+  Output,
+} from '@angular/core';
+
+import { CopyService } from './copy-service/copy.service';
+
+export abstract class CopyBase {
+  @Input('copyBtnAriaLabel') copyBtnAriaLabel: string;
+  @Input('copyValue') copyValue: string = 'Missing \'copyValue\' @Input property';
+  @Input('tooltip') tooltip: string;
+  @Input('tooltipPlacement') tooltipPlacement: string = 'top';
```
dlabrecq: Needs typedoc comments

src/app/copy/copy-base.ts
```diff
@@ -0,0 +1,59 @@
...
+  public get recentlyCopied(): boolean {
+    return this._recentlyCopied;
+  }
+
+  /**
+   * Set the flag indicating copy action has just happened
+   *
+   * @param {boolean} copied True when copy action has been triggered
+   */
+  public set recentlyCopied(copied: boolean) {
```
dlabrecq: Does this setter need to be public? I'm not sure it's used by the component itself, so perhaps it can be removed?

src/app/copy/copy-base.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Default constructor
+   */
+  constructor(protected copyService: CopyService) {}
+
+  /**
+   * Copy value to the user's system clipboard
+   */
+  copyValueToClipboard(accessibleName: string): void {
```
dlabrecq: Can this method be marked protected?

dlabrecq (changes requested):

src/app/copy/copy-base.ts
seanforyou23: So when I use protected, I get the aot error

> ERROR in ng:///workingrepos/seanforyou23-patternfly-ng/src/app/copy/inline-copy/inline-copy.component.html (13,5): Property 'copyValueToClipboard' is protected and only accessible within class 'InlineCopyComponent' and its subclasses.

The default seemed to work ok so I was going with that. I'm looking into it more now though, feel free to make a suggestion if this is a familiar error

src/app/copy/copy-base.ts
seanforyou23: Similar to copyValueToClipboard, if I change this to private/protected I get build errors. It's possible I'm just missing something so I'll take another look.

src/app/copy/inline-copy/inline-copy-config.ts
seanforyou23: I can!

src/app/copy/copy-base.ts
dlabrecq: Instead of calling copyValueToClipboard directly within the HTML template, you could do something more like this:

copyValue(): void {
  this.copyValueToClipboard(this.copyBtnAriaLabel);
}

src/app/copy/copy-base.ts
dlabrecq: It appears that only CopyBase sets the recentlyCopied property. Instead of invoking this:

this.recentlyCopied = true;

You can set the property like so:

this._recentlyCopied = true;

That way, you can get rid of the public setter.

src/app/copy/block-copy/block-copy-config.ts
seanforyou23: So the thought here was to specify these shared values as part of a config class to ensure consistency in the names/types of component inputs. It was nice to be able to import this class in the tests and example files for both copy components as it sort of guided the development experience. I suppose the value here isn't has big as I first thought. I'll remove this config class and redefine these inputs for each of the copy components.

dlabrecq: That's an interesting idea and I did give that some thought. However, considering we have the same import properties on the component, it's extra APIs and typedoc to maintain. The fewer public APIs the better -- more difficult to deprecate later.

Going forward, I'd like to deprecate the config objects we have today in favor of component imports. It's easier for developers to bind properties directly to component tags rather than setup a config object in the view controller.

That said, we will likely keep some config objects for third party libraries. For example, using a config object to specify only C3 properties may still be helpful for chart components.

src/app/copy/copy-base.ts
seanforyou23: Oh! Geez I feel silly now :) For some reason I was thinking the getters and setters were a package deal and I needed to _use_ the setter. Was definitely overcomplicating this, good call!

src/app/copy/copy-base.ts
seanforyou23: Ah ok, I see what you mean. I was actually considering doing this before but with a new build I'm realizing that error was just a one-off. Thanks for explaining that alternative though.

seanforyou23: Oh, I see in your other comment there is preference to make these protected, so yea I'll be taking this advice after all.

src/app/copy/block-copy/block-copy-config.ts
seanforyou23: Thanks for the clarification, that sounds like a great strategy.

seanforyou23:
>I did notice that when I have focus on a Copy button and hit Enter, the focus isn't maintained on the Copy button. I would expect that focus to remain on the action I selected unless there's a reason to shift focus

@dlabrecq any chance you know what gives on this? Been trying to figure this out but no luck so far. I see this working as expected with buttons on several other widgets, for example

https://rawgit.com/patternfly/patternfly-ng/master-dist/dist-demo/#/sparkline and
https://rawgit.com/patternfly/patternfly-ng/master-dist/dist-demo/#/emptystate

But in this component, focus falls into the abyss once you press the enter key. Setup looks the same for the event handlers.. I've tried all the obvious things like returning false, preventing default, etc.

I thought it may be related to the button's html being replaced but I get the same behavior even with static html.

I could programmatically re-place the focus on the button, but this seems out of the way and I'm hoping to find a more natural solution. Let me know if you have any ideas, I'll keep looking for a solution in the mean time. Thanks!

dlabrecq:
I changed your code to use a checkbox instead of a button and saw the same results.

My guess is the HTML changes in a way that the original element (with focus) no longer exists? Looking at the code I see two icons.

```
<i
      class="fa fa-clipboard"
      aria-hidden="true"
      *ngIf="!recentlyCopied"></i>
    <i
      class="fa fa-check"
      aria-hidden="true"
      *ngIf="recentlyCopied"></i>
```

Instead of hiding and showing two different icons, perhaps you can just apply the different CSS selectors to the same icon? Probably a cleaner way to apply different styles, anyway. For example:

```
<i class="fa" aria-hidden="true" 
    [ngClass]="'{'fa-check': recentlyCopied, 'fa-clipboard': !recentlyCopied}"/>
```

dlabrecq:
BTW, I noticed this CobyBase code emits a string. This should probably be a typed object with the name as a property. Otherwise, folks may need to parse a string to figure out which component generated the event.

this.copiedToClipboard.emit(`${accessibleName} copied!`);

dlabrecq:
Looking at the CopyService and believe the issue is related to creating a new HTML element in order to perform the copy. Specifically, we're calling this:

`textarea.select();`

When I comment this out, focus remains. Unfortunately, that is required to copy; thus, we will have to reset focus manually.

dlabrecq:
We should add a warning to ServiceCopy.copy() method to indicate that we are setting focus here. Alternatively, we can obtain the element that has current focus in order to restore it later. For example, we should be able to use document.activeElement or document.querySelector( ':focus' )

See: https://stackoverflow.com/questions/497094/how-do-i-find-out-which-dom-element-has-the-focus

src/app/copy/copy-base.ts
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Copy value to the user's system clipboard
+   * @param {string} accessibleName An accessible name used to describe the component
+   */
+  protected copyValueToClipboard(accessibleName: string): void {
+    let result = this.copyService.copy(this.copyValue);
+    if (result) {
+      this.copiedToClipboard.emit({
+        name: accessibleName,
+        msg: `${accessibleName} copied`
```
dlabrecq, line 76: Can this event contain the copied value instead? I don't think the msg string will be very useful and it's not localizable.

Instead of providing the aria label here, I'm thinking that the copied value will be more appropriate. Unless we also have a component ID, name, etc. the aria label itself really doesn't identify which component generated the event. That is, considering the tooltip/aria label are optional, the event could be empty.

src/app/copy/copy-base.ts
```diff
@@ -0,0 +1,84 @@
...
+   * Copy button aria label (announced to screen readers)
+   * @type {string}
+   */
+  @Input('copyBtnAriaLabel') copyBtnAriaLabel: string;
+
+  /**
+   * The text node to be copied to the users clipboard
+   * @type {string}
+   */
+  @Input('copyValue') copyValue: string = 'Missing \'copyValue\' @Input property';
```
dlabrecq, line 31: Can this be an empty string or left undefined? I'd prefer not to hard code strings as default properties. An application would have to parse this string in order to interpret its meaning. Testing the copy value for an empty string or undefined would be easier to determine if a value was obtained.

src/app/copy/block-copy/block-copy.component.ts
```diff
@@ -0,0 +1,73 @@
...
+  Input,
+  ViewEncapsulation
+} from '@angular/core';
+import { uniqueId } from 'lodash';
+
+import { CopyBase } from '../copy-base';
+
+import { CopyService } from '../copy-service/copy.service';
+
+@Component({
```
dlabrecq, line 12: Needs typedoc comments for the component itself. Can be very simple with basic import examples. See basic card for an example.

```
/**
 * Card component
 *
 * For customization, use the templates named headerTemplate and footerTemplate.
 *
 * Usage:
 * <br/><code>import { BasicCardModule } from 'patternfly-ng/card';</code>
 *
 * Or:
 * <br/><code>import { BasicCardModule } from 'patternfly-ng';</code>
 */
```

src/app/copy/block-copy/example/block-copy-basic-example.component.ts
```diff
@@ -0,0 +1,28 @@
...
+    buttonLabel: 'Copy JSON',
+    tooltip: 'Example Swagger JSON Snippet',
+    expandBtnAriaLabel: 'Toggle Swagger JSON',
+    copyBtnAriaLabel: 'Copy Swagger JSON'
```
dlabrecq, line 21: It appears that buttonLabel and copyBtnAriaLabel are used for the same HTML element. Can they be named similarly? For example, buttonLabel and buttonAriaLabel or copyBtnLabel and copyBtnAriaLabel.

Whatever you choose, please make inline-copy the same.

src/app/copy/copy-service/copy.service.module.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { CopyService } from './copy.service';
+
+@NgModule({
+  imports: [
+    CommonModule
+  ],
+  providers: [CopyService]
+})
+export class CopyServiceModule {}
```
dlabrecq, line 12: Please remove this module. We're not declaring a component or directive, so a module isn't necessary. It's already exported via the index.ts file.

FYI, the notification service no longer has a module -- it's been deprecated.

src/app/copy/copy-base.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { CopyService } from './copy-service/copy.service';
+
+  /**
+   * A standard structure for result of copied to clipboard action
+   */
+export interface CopiedMsg {
+  name: string;
+  msg: string;
+}
```
dlabrecq, line 15: CopiedMsg should be exported via the index.js file. 

I'd also like to rename this object as CopyEvent, located in its own file, with typedoc comments. This would be similar to other component events.

src/app/copy/inline-copy/inline-copy.component.ts
```diff
@@ -0,0 +1,32 @@
+import {
+  Component,
+  ViewEncapsulation
+} from '@angular/core';
+
+import { CopyBase } from '../copy-base';
+
+import { CopyService } from '../copy-service/copy.service';
+
+@Component({
```
dlabrecq, line 10: Needs typedoc comments for the component itself. See basic card for an example.

dlabrecq (changes requested):
Looking good. Just minor comments

src/app/copy/copy-service/copy.service.ts
```diff
@@ -0,0 +1,73 @@
...
+    textarea.style.top = '-100px';
+    textarea.style.left = '-100px';
+    textarea.style.opacity = '0';
+    textarea.value = value;
+
+    this.dom.body.appendChild(textarea);
+
+    textarea.select();
+
+    if (!!triggerElement) {
```
dlabrecq, line 39: Using `!!` is an obscure way to perform a type conversion. Please use one of the following approaches instead.

if (triggerElement !== undefined) {...} or
if (triggerElement) {...}

dlabrecq (approved):
Will address lingering code review comments in a new PR

PR #352, opened 2018-05-17
Open peer dependencies to include angular 6

dlabrecq:
Tested using the latest angular-cli and built with `ng --prod`. My Angular 6 based app built and ran patternfly-ng components without error.

Note that angular-cli isn't without its own issues. I encountered this problem along the way, but was able to apply the suggested workaround.
https://github.com/aws/aws-amplify/issues/678

The angular-cli application also required rxjs-compat 6.1.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,9 +42,15 @@
     "angular 2",
     "angular4",
     "angular 4",
+    "angular5",
+    "angular 5",
+    "angular6",
+    "angular 6",
     "ng",
     "ng2",
-    "ng4"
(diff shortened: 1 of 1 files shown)


dlabrecq:
Although initial testing showed the current version of patternfly-ng runs with Angular 6, I'm closing this after the challenges noted in https://github.com/patternfly/patternfly-ng/issues/356. We will likely create a new major version of patternfly-ng that's built with Angular 6 instead.

PR #355, opened 2018-05-18
Support rxjs v6

mprahl:
Angular v6 started requiring rxjs v6 which has changed import locations. ngx-bootstrap v2 does not support rxjs v6, so as a work-around, we can install rxjs-compat to get Patternfly to work with Angular v6.

When you feel comfortable, you could alternatively upgrade ngx-bootstrap to v3. Here is the backstory on that:
https://github.com/valor-software/ngx-bootstrap/issues/4308

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,8 @@
     "angular-tree-component": "^6.0.0 || ^7.0.0",
     "c3": "^0.4.15",
     "ng2-dragula": "^1.5.0",
-    "ngx-bootstrap": "^1.8.0 || ^2.0.0"
+    "ngx-bootstrap": "^1.8.0 || ^2.0.0",
+    "rxjs-compat": "^6.1.0"
   },
   "devDependencies": {
     "@angular/animations": "^4.0.1",


package.json
```diff
@@ -75,7 +75,8 @@
     "angular-tree-component": "^6.0.0 || ^7.0.0",
     "c3": "^0.4.15",
     "ng2-dragula": "^1.5.0",
-    "ngx-bootstrap": "^1.8.0 || ^2.0.0"
+    "ngx-bootstrap": "^1.8.0 || ^2.0.0",
+    "rxjs-compat": "^6.1.0"
```
dlabrecq, line 79: Considering it's the application that performs the transpile (patternfly-ng doesn't use rxjs-compat itself), I'm thinking it should be the application that includes rxjs-compat. That way, an unused package is not installed by default for existing Angular 4 and 5 applications.

Since there are breaking changes with rxjs 6 (the reason for rxjs-compat), and the numerous challenges noted in #356, I'd like to create a new major version of patternfly-ng built with Angular 6 instead. Then, we can update to the latest versions of ngx-bootstrap 3.0.0, ngx-datatable 13.0.0, rxjs 6, and webpack 4.

Note that patternfly-ng currently builds with Angular 4 in order to be compatible with 4 and 5. However, the current version of patternfly-ng cannot upgrade to ngx-bootstrap 3.0.0 and ngx-datatable 13.0.0 because those packages are built with Angular 6. In order to update those packages, patternfly-ng must also build with Angular 6, which will be a breaking change.

dlabrecq: See: https://github.com/patternfly/patternfly-ng/pull/359

Issue #348, opened 2018-05-16
Lists do not generate element ids


kpiwko:
It would very much simplify navigation in the list by automated tools (e.g. Selenium) if list related components would be able to add `id` attribute to each element.

For instance here:
https://github.com/patternfly/patternfly-ng/blob/a23d41d04b8ed607ddcdc6d3d69a873966edeb53/src/app/list/basic-list/list.component.html#L63

There could be something that adds `id="item.id"` if any such value has been defined.

dlabrecq:
The selected state is maintained on the provided item itself, so it's not clear why you need an ID on the checkbox? 

If you're looking to test the current selections, you should already have them. The component sets the selected property on the array of objects (items) provided. 

Note that the component also emits an event whenever a selection is made, which includes an array of current selections.

kpiwko:
The main benefit of having an id attribute is that id selector is the quickest and the most stable across browsers - this is a huge benefit for blackbox tests. Id is used to navigate at the page - it is not only about which item is selected but also about ability to click on different checkbox.

https://blog.mozilla.org/fxtesteng/2013/09/26/writing-reliable-locators-for-selenium-and-webdriver-tests/

dlabrecq:
Checkboxes are not shown by default, so while that would help your particular Selenium test case, it may not benefit others using a basic list. We may need to consider placing an ID on the row itself so this can work for all list states?

kpiwko:
You are right, I haven't realized that there are lists without checkboxes.

An id on a row itself would help as well, it would need a CSS selector to hit checkbox but still it would vastly simply selectors.

dlabrecq:
You could probably use something more like input[@type='checkbox']. That could also used to obtain all the checkboxes in the page without IDs.

https://stackoverflow.com/questions/14682763/how-to-select-checkboxes-using-selenium-java-webdriver

kpiwko:
Yes, we could do something like `input[type='checkbox']`. However, since Patternfly does not render **value** nor **name** attribute at input in lists, navigation from there to find the right checkbox would be very complex. TBH, I'm not sure, is `value="item.selected"` for `input`s rendering in PF lists bug or feature?

With an id on row, I could use something like `#myrowid input[type='checkbox']` and I'm all set.

dlabrecq:
Thanks for pointing that out -- that's a bug. The value attribute has no effect since the actual value is stored internally via Angular's ngModel.

[...]

PR #358, opened 2018-05-22
feat(list-pf): radio button to support id

corinnekrych:


corinnekrych:
fixes https://github.com/patternfly/patternfly-ng/issues/348

kpiwko:
Hello @corinnekrych, this change collides with #357 where value has been removed.

Also, based on the discussion we had at #348, your change is not complete - first, we want to have ids at input of type checkbox as well and second based on what @dlabrecq said, it would be more flexible if any row would contain an id attribute - so we could navigate rows in pf-lists in general.

corinnekrych:
Closing it then! I guess for now you could use the sibling selector

PR #359, opened 2018-05-22
Angular 6 support

dlabrecq:
DO NOT MERGE: BREAKING CHANGE

Updated to Angular 6, ngx-bootstrap 3.0.0, ngx-datatable 13.0.0, rxjs 6, gulp 4, and webpack 4

Dropped the tree-list component due to additional issues with mobx autorun introduced by angular-tree-component.
See: https://github.com/patternfly/patternfly-ng/issues/381

~~Note: Karma won't run without typescript 2.8.1, while Angular 6 requires 2.7.2.
See: https://stackoverflow.com/questions/49468616/cannot-read-property-externalmoduleindicator-of-undefined-while-running-unit-t~~


seanforyou23:
>Note: Karma won't run without typescript 2.8.1, while Angular 6 requires 2.7.2.

I found that by replacing `awesome-typescript-loader` with `ts-loader`, karma is able to start while using typescript 2.7.2

dlabrecq:
Replaced awesome-typescript-loader with ts-loader -- unit tests run as expected now.

dlabrecq:
Replaced extract-text-webpack-plugin with mini-css-extract-plugin. The extract-text-webpack-plugin simply does not work with webpack 4 and prevents the demo from being able to watch for code changes.

The mini-css-extract-plugin is compatible with webpack 4 and restores the demo's ability to watch for updates. Unfortunately, I had to temporarily disable AOT for the demo because mini-css-extract-plugin has an issue with AngularCompilerPlugin.

See: https://github.com/webpack-contrib/mini-css-extract-plugin/issues/186

dlabrecq:
Updated to gulp 4 and added an AOT build. This means the build will be much quicker because we don't need to run the demo just to test AOT.

PR #360, opened 2018-05-23
List: add unique element IDs

dlabrecq:
This adds unique element IDs for each row, checkbox, and radio button. Default IDs can be overridden by providing an id for the pfng-list tag.

Example
https://rawgit.com/dlabrecq/patternfly-ng/list-ids-dist/dist-demo/#/list

Fixes: https://github.com/patternfly/patternfly-ng/issues/348

screen shot 2018-05-22 at 8 49 50 pm


joshuawilson:
nice!

corinnekrych (dismissed):

Issue #322, opened 2018-02-14
Dropdown not working on Vertical Navigation Demo


dhilst:
Hi, 
I'm was trying the Vertical Navigation Demo and find out that the top right dropdowns (help, user) are not showing. I tried at the live demo and got the same result. I can see the buttons but it's not dropping down.

Am I doing something wrong?

https://rawgit.com/patternfly/patternfly-ng/master-dist/dist-demo/#/verticalnavigation

chouseknecht:
@dhilst 

Ran into the same issue. The API site example seems to be incomplete. You have to import the `BsDropdownModule` from `ngx-bootstrap`, like so:

```
import { BrowserModule }        from '@angular/platform-browser';
import { NgModule }             from '@angular/core';
import { AppComponent }         from './app.component';

import { NavigationModule }     from 'patternfly-ng';
import { AppRoutingModule }     from './/app-routing.module';
import { HomeComponent }        from './home/home.component';

import { BsDropdownModule }     from 'ngx-bootstrap';

@NgModule({
    declarations: [
        AppComponent,
        HomeComponent
    ],
    imports: [
        BrowserModule,
        NavigationModule,
        AppRoutingModule,
        BsDropdownModule.forRoot()
    ],
    providers: [],
    bootstrap: [AppComponent]
})
export class AppModule { }
```


Next, add `*dropdownMenu`, `dropdown` and `dropdownToggle` to the HTML, as in the following:

```
<div id="verticalNavLayout" class="layout-pf layout-pf-fixed faux-layout" style="background-color: white;">
    <pfng-vertical-navigation
        [brandAlt]="'Galaxy'"
        [contentContainer]="contentContainer"
        [items]="navigationItems"
        [persistentSecondary]="false"
        [pinnableMenus]="true"
        [showBadges]="true"
        [showIcons]="true"
        [updateActiveItemsOnClick]="true">
        <div>
            <ul class="nav navbar-nav navbar-right navbar-iconic">
                <li class="dropdown" dropdown>
                    <a class="dropdown-toggle nav-item-iconic" id="dropdownMenu1" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true" dropdownToggle>
                        <span title="Help" class="fa pficon-help"></span>
                        <span class="caret"></span>
                    </a>
                    <ul *dropdownMenu class="dropdown-menu" aria-labelledby="dropdownMenu1">
                        <li><a href="#">Action</a></li>
[...]

PR #363, opened 2018-05-23
fix(demo): vertical navigation example dropdowns

jschuler:
Fixing the dropdowns as suggested in https://github.com/patternfly/patternfly-ng/issues/322


dlabrecq (approved):
Looks good. Thanks for the contribution!

Issue #317, opened 2018-02-08
List item content area clickability


seanforyou23:
We found a discrepancy in the list component behavior between angular-patternfly and patternfly-ng.

In the below links, select the "simple expansion"/"show expanding rows" option for the list component example, and notice that in angular-patternfly clicking anywhere inside of the list item content will expand the container, revealing previously hidden content.

In patternfly-ng, users must click the (much smaller) little arrow for a given item to expand its contents.

Further, the hover state for the cursor is not reflected in patternfly-ng so it's not so obvious where you need to click to expand the content - or that clicking an item will do anything at all! Changing the cursor style to "pointer" while hovering an expandable item would help here.

http://www.patternfly.org/angular-patternfly/#/api/patternfly.views.component:pfListView
http://www.patternfly.org/patternfly-ng/#/list

This was discovered in an unrelated issue https://github.com/syndesisio/syndesis/issues/1467

PR #369, opened 2018-05-29
fix(list): make more areas clickable for expansion and add hover styles

jschuler:
Addresses https://github.com/patternfly/patternfly-ng/issues/317

The border of the list item as well as the area around the left chevron is now also clickable if expansion is enabled.
Also the chevron now changes color on hover over (tested for basic list expansion, compound expansion, tree list).

### DIFF
--- a/src/app/list/basic-list/list.component.ts
+++ b/src/app/list/basic-list/list.component.ts
@@ -152,7 +152,17 @@ export class ListComponent extends ListBase implements DoCheck, OnInit {
     item.expanded = false;
   }
 
-  private toggleExpandArea(item: any): void {
+  /**
+   * Toggles the list item expansion
+   * 
+   * @param {MouseEvent} $event The event emitted when an item has been clicked
+   * @param {Object} item The object associated with the current row
+   */
+  private toggleExpandArea($event: any, item: any): void {
+    // Do nothing if item expansion is disabled
+    if (!this.config.useExpandItems) return;
+    // Do not trigger for child items, only on the DOM element to which the event is attached
+    if ($event.target !== $event.currentTarget) return;
     // Item may already be open due to compound expansion
     if (item.expanded && item.expandId !== undefined) {
       item.expandId = undefined;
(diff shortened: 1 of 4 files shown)


src/app/list/basic-list/list.component.ts
```diff
@@ -152,7 +152,17 @@ export class ListComponent extends ListBase implements DoCheck, OnInit {
...
   }
 
-  private toggleExpandArea(item: any): void {
+  /**
+   * Toggles the list item expansion
+   * 
+   * @param {MouseEvent} $event The event emitted when an item has been clicked
+   * @param {Object} item The object associated with the current row
+   */
+  private toggleExpandArea($event: any, item: any): void {
```
dlabrecq: I believe the event type should be `MouseEvent`. We don't want to use type `any` when possible -- `item` is an exception.

src/app/list/basic-list/list.component.ts
```diff
@@ -152,7 +152,17 @@ export class ListComponent extends ListBase implements DoCheck, OnInit {
...
+    // Do nothing if item expansion is disabled
+    if (!this.config.useExpandItems) return;
+    // Do not trigger for child items, only on the DOM element to which the event is attached
+    if ($event.target !== $event.currentTarget) return;
```
dlabrecq: Please use brackets for the if-statements.

dlabrecq:
It appears you have setup Travis successfully, so it would be great to include an example link for other reviewers. A screenshot can also be helpful.

Your rawgit URL (for this example) is:
https://rawgit.com/jschuler/patternfly-ng/list-item-toggle-expand-dist/dist-demo/#/list

dlabrecq (changes requested):
I was able to build and test locally. Looks great!

I'm glad you were able to address the actions area. Clicking on the action buttons still works as expected.

Just a couple very small nits

jschuler:
Thanks @dlabrecq, made those changes

dlabrecq (approved):
Looks good !

Issue #333, opened 2018-04-17
List component "simple expansion" styling issue


seanforyou23:
Notice in the [design overview for list](http://www.patternfly.org/pattern-library/content-views/list-view/#simple-expansion), the content container for the expanded list item has a border and dropshadow to distinguish it from the other page elements.

But then in the pf-ng component, there is no border or dropshadow around the expanded content container (you'll need to select the "Simple Expansion" checkbox in the [Basic/Compound Expansion list demos](http://www.patternfly.org/patternfly-ng/#/list) to see it in action.

Finally, here's an [example implementation](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/list-view-simple-expansion.html) that showcases the proper html/css used to achieve this. On the surface it looks like the issue is misalignment of the html/css.

dlabrecq:
Not certain if the drop shadow is a more recent addition, but it's an easy fix. We simply need to apply the 'active' style when the row is expanded, like so:

`<div class="list-pf-item active"...`

jschuler:
Fixed with:
https://github.com/patternfly/patternfly-ng/pull/370
https://github.com/patternfly/patternfly/pull/1070

PR #370, opened 2018-05-31
fix(list): set list item active when expanded

jschuler:
Addresses issue in https://github.com/patternfly/patternfly-ng/issues/333

Change:
When a list item is expanded, the correct boolean variable is used to set the active class on list items.

Note:
There will be another PR to patternfly-core to address the .list-pf-item.active class such as background-color and box-shadow.

Travis build:
https://rawgit.com/jschuler/patternfly-ng/list-item-active-dist/dist-demo/#/list


dlabrecq (approved):

dlabrecq:
Thanks for the fix!

PR #374, opened 2018-06-07
Angular6 .angular-cli.json --> angular.json

shawndwells:
README used old angular-cli.json vs new angular.json. This PR fixes that.


dlabrecq (approved):
These changes look good, but the current version of patternfly-ng still runs Angular 4/5. We should hold this and merge after PR https://github.com/patternfly/patternfly-ng/pull/359

Thank you for the update!

PR #380, opened 2018-06-14
fix(toolbar): give accessible name to sort and more actions button

seanforyou23:
This PR makes some small but impactful changes to accessible text found in the toolbar. The Sort and More Actions buttons where not displaying in the rotor menu at all, but with this PR they are properly labelled. This also cleans up the various icons that simply needed to be hidden.

hide icons from screen readers
use accessible names
ensure accessible names appear for cursor navigation or rotor

Demo should be available from link below once travis finishes
https://rawgit.com/seanforyou23/patternfly-ng/a11y-adjustments-dist/dist-demo/

fixed

### DIFF
--- a/src/app/action/action-config.ts
+++ b/src/app/action/action-config.ts
@@ -11,6 +11,11 @@ export class ActionConfig {
   moreActions?: Action[];
 
   /**
+   * Text announced to screen readers for the action config button
+   */
+  moreActionsAriaLabel?: string;
+  
+  /**
    * Set to true to disable secondary actions
    */
   moreActionsDisabled: boolean;
@@ -29,9 +34,4 @@ export class ActionConfig {
    * List of primary actions (e.g., for buttons)
    */
   primaryActions: Action[];
-
-  /**
-   * Text announced to screen readers for the action config button
-   */
-  ariaLabel?: string;
 }
--- a/src/app/action/action-config.ts
+++ b/src/app/action/action-config.ts
@@ -29,4 +29,9 @@ export class ActionConfig {
    * List of primary actions (e.g., for buttons)
    */
   primaryActions: Action[];
+
+  /**
+   * Text announced to screen readers for the action config button
+   */
+  ariaLabel?: string;
 }
--- a/src/app/action/action.component.html
+++ b/src/app/action/action.component.html
@@ -23,7 +23,8 @@
      *ngIf="config.moreActions?.length > 0">
   <button class="btn btn-link dropdown-toggle" type="button" dropdownToggle
           [ngClass]="{'disabled': config.moreActionsDisabled}"
-          (click)="initMoreActionsDropup($event)">
+          (click)="initMoreActionsDropup($event)"
+          aria-label="More Actions">
     <span class="fa fa-ellipsis-v"></span>
   </button>
   <ul class="dropdown-menu-right dropdown-menu" aria-labelledby="dropdownKebab" *dropdownMenu>
--- a/src/app/action/action.component.html
+++ b/src/app/action/action.component.html
@@ -24,7 +24,7 @@
   <button class="btn btn-link dropdown-toggle" type="button" dropdownToggle
           [ngClass]="{'disabled': config.moreActionsDisabled}"
           (click)="initMoreActionsDropup($event)"
-          [attr.aria-label]="config.moreActionsAriaLabel || 'More Actions'">
+          [attr.aria-label]="config.moreActionsAriaLabel">
     <span class="fa fa-ellipsis-v"></span>
   </button>
   <ul class="dropdown-menu-right dropdown-menu" aria-labelledby="dropdownKebab" *dropdownMenu>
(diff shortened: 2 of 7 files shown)


src/app/action/action.component.html
```diff
@@ -23,7 +23,8 @@
      *ngIf="config.moreActions?.length > 0">
   <button class="btn btn-link dropdown-toggle" type="button" dropdownToggle
           [ngClass]="{'disabled': config.moreActionsDisabled}"
-          (click)="initMoreActionsDropup($event)">
+          (click)="initMoreActionsDropup($event)"
+          aria-label="More Actions">
```
dlabrecq: Perhaps we should allow this to be set via an ariaLabel property in the Action object used by ActionConfig. (Unless using the Action tooltip is sufficient?) As a default value, we could use the Action tooltip property, if necessary?

src/app/toolbar/toolbar.component.html
```diff
@@ -26,9 +26,12 @@
              *ngIf="viewTemplate !== undefined || (config.views)">
           <ng-template [ngTemplateOutlet]="viewTemplate" [ngTemplateOutletContext]="{}"></ng-template>
           <span *ngIf="config.views">
-            <button *ngFor="let view of config.views" class="btn btn-link"
+            <button *ngFor="let view of config.views"
+                    class="btn btn-link"
                     [ngClass]="{'active': isViewSelected(view), 'disabled': view.disabled === true}"
-                    title={{view.tooltip}}  (click)="viewSelected(view)">
+                    [attr.aria-label]="view.tooltip"
+                    title={{view.tooltip}}
```
dlabrecq: Perhaps we should add a separate ariaLabel property in ToolbarView? We could use the ToolbarView tooltip as a default, if necessary?

Whatever we do, lets try to be consistent with the copy component. I believe we're introducing separate ariaLabel properties there, but we can back that out and use the tooltip for all components if that's sufficient?

dlabrecq (changes requested):

src/app/action/action.component.html
seanforyou23: By Action object, I assume you're talking about actionConfig in the toolbar example component?  I added ariaLabel as an optional property on ActionConfig and set the value to "More Actions" if ariaLabel isn't specified. There wasn't already a tooltip property in ActionConfig so I didn't see sense in adding one just to set it as a default for ariaLabel, since the tooltip would be optional also.

The tooltip for the individual actions is a different thing though, since they are specific to each action. The ariaLabel here is really describing that entire "More Actions" kabab as a whole so it seems like adding a dedicated property for this, with a text fallback, is a good thing to do.

Thoughts?

src/app/toolbar/toolbar.component.html
seanforyou23: I like the idea of first introducing ariaLabel properties so there is at least the potential for customizing them. I also like the idea of using tooltip as a default in case users do not specify an ariaLabel since they both are intended to describe the element. Seems like that's kinda the best of both worlds.

Sometimes a tooltip is not desired though, so in those scenarios I was thinking setting a text default might be a happy middle ground. In this way there's always a value supplied so screen reader experience doesn't appear broken, and it doesn't necessarily add any overhead to the one using the component.

@jgiardino thoughts on using tooltip text values as backup/defaults for aria-label if they are not supplied when instantiating a component? My only real concern is that they are all optional properties, so unless we specify a text default at some point, there's always the possibility that screen reader user winds up with a poor experience.

jgiardino: Can you clarify what controls you're considering doing this for?

When you say `title` is this the tooltip component, or the `title` attribute?

src/app/action/action-config.ts
```diff
@@ -29,4 +29,9 @@ export class ActionConfig {
    * List of primary actions (e.g., for buttons)
    */
   primaryActions: Action[];
+
+  /**
+   * Text announced to screen readers for the action config button
+   */
+  ariaLabel?: string;
```
dlabrecq: I was expecting this to be on the Action object, along side the tooltip and title properties. (That would allow a unique property for each button and kebab defined by ActionConfig.) However, it appears that this property is only used for the kebab (i.e., the more actions menu and not any primary actions)? If so, please rename it as moreActionsAriaLabel so it's clear what component the property will be applied to. That is, it's not being a

src/app/action/action.component.html
```diff
@@ -24,7 +24,7 @@
   <button class="btn btn-link dropdown-toggle" type="button" dropdownToggle
           [ngClass]="{'disabled': config.moreActionsDisabled}"
           (click)="initMoreActionsDropup($event)"
-          aria-label="More Actions">
+          [attr.aria-label]="config.ariaLabel || 'More Actions'">
```
dlabrecq: I don't believe we should provide the "More Actions" string by default. Unless we commit to localizing our components, hard coded strings don't work well across applications. This would break the UI as soon as someone changes the preferred browser locale. We don't provide default button labels; for example, even though it would look silly without one.

That said, if we had a moreActionsTooltip property, perhaps that could be used as a default.

dlabrecq (changes requested):

dlabrecq (approved):

PR #396, opened 2018-06-21
vertical-nav: close secondary and tertiary menus upon leaf node click

### LINKED ISSUE #372: VerticalNavigationComponent doesn't close secondary on click
The PatternFly VerticalNavigation shows that when you click an item in a secondary menu then the menu will immediately close.  I can't find a way to make this happen with the component.

dlabrecq:
This change closes the secondary and tertiary menus upon leaf node click. Otherwise, the user must hover or click outside the navigation area, then wait a couple seconds for the menu to dismiss.

Fixes https://github.com/patternfly/patternfly-ng/issues/372

FYI @ssilvert


ssilvert:
@dlabrecq Thanks!

PR #401, opened 2018-06-26
applies font awesome's fixed-width class to the 'angle' icon in the basic list component

brimwd:
https://fontawesome.com/v4.7.0/examples/#fixed-width

This makes the font awesome icons a uniform width, so when the icon changes from `angle-right` to `angle-down` the adjacent label doesn't shift position.


dlabrecq (approved):
Tried this out locally and it looks good to me. Thanks for the fix!

PR #435, opened 2018-07-31
fix(list): allow setting a custom trackBy function for the underlying ngFor directive

riccardo-forina:
It is now possible to set a custom trackBy function to be used in the underlying ngFor directive, which by default tracks changes in the items by identity.

This is useful in scenarios where the items passed to the component are periodically polled, and components with state (eg. a `pfng-action` component) are used in the item's template.

This should help fixing syndesisio/syndesis/issues/3121 where we have exactly this problem.

### DIFF
--- a/src/app/list/basic-list/example/list-basic-example.component.ts
+++ b/src/app/list/basic-list/example/list-basic-example.component.ts
@@ -27,7 +27,6 @@ export class ListBasicExampleComponent implements OnInit {
   itemsAvailable: boolean = true;
   listConfig: ListConfig;
   selectType: string = 'checkbox';
-  setItemInterval: number;
 
   constructor() {
   }
@@ -115,6 +114,7 @@ export class ListBasicExampleComponent implements OnInit {
       nodeCount: 10,
       imageCount: 8
     }];
+    this.items = cloneDeep(this.allItems);
 
     this.emptyStateConfig = {
       actions: {
@@ -159,20 +159,9 @@ export class ListBasicExampleComponent implements OnInit {
       showRadioButton: false,
       useExpandItems: false
     } as ListConfig;
-
-    this.setItemInterval = setInterval(() => this.setItems(), 1000);
-    this.setItems();
   }
 
   ngDoCheck(): void {
-  }
-
-  ngOnDestroy(): void {
-    clearInterval(this.setItemInterval);
-  }
-
-  setItems(): void {
-    this.items = cloneDeep(this.allItems);
   }
 
   /**
@@ -259,17 +248,6 @@ export class ListBasicExampleComponent implements OnInit {
       actionConfig.moreActionsVisible = false;
     }
     return actionConfig;
-  }
-
-  /**
-   * Get the tracking id to use for each row
-   *
-   * @param index The current row index
-   * @param item The current row item
-   * @returns number
-   */
-  getTrackBy(index: number, item: any): any {
-    return index;
   }
 
   // Actions
--- a/src/app/list/basic-list/example/list-basic-example.component.ts
+++ b/src/app/list/basic-list/example/list-basic-example.component.ts
@@ -27,6 +27,7 @@ export class ListBasicExampleComponent implements OnInit {
   itemsAvailable: boolean = true;
   listConfig: ListConfig;
   selectType: string = 'checkbox';
+  setItemInterval: number;
 
   constructor() {
   }
@@ -114,7 +115,6 @@ export class ListBasicExampleComponent implements OnInit {
       nodeCount: 10,
       imageCount: 8
     }];
-    this.items = cloneDeep(this.allItems);
 
     this.emptyStateConfig = {
       actions: {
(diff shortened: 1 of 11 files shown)


src/app/list/basic-list/list.component.ts
```diff
@@ -162,6 +167,10 @@ export class ListComponent extends ListBase implements DoCheck, OnInit {
     return result + '-' + suffix + index;
   }
 
+  protected getTrackBy(index: number, item: any): string {
+    return `track-${index}`;
+  }
```
dlabrecq: This function appears to be unused.

src/app/list/basic-list/list.component.ts
```diff
@@ -6,7 +6,7 @@ import {
   Input,
   OnInit,
   Output,
-  TemplateRef,
+  TemplateRef, TrackByFunction,
```
dlabrecq: own line?

src/app/list/basic-list/list.component.spec.ts
```diff
@@ -294,3 +296,137 @@ describe('List component - ', () => {
...
+    this.config = {
+      dblClick: false,
+      emptyStateConfig: this.emptyStateConfig,
+      multiSelect: false,
+      selectItems: false,
+      selectionMatchProp: 'name',
+      showCheckbox: false,
+      useExpandingRows: false
+    } as ListConfig;
+  }
```
dlabrecq, line 301: Much of this config is duplicated, top of the spec. Can we reuse any of that?

src/app/list/basic-list/example/list-basic-example.component.ts
```diff
@@ -250,6 +261,17 @@ export class ListBasicExampleComponent implements OnInit {
...
+   * Get the tracking id to use for each row
+   *
+   * @param index The current row index
+   * @param item The current row item
+   * @returns number
+   */
+  getTrackBy(index: number, item: any): any {
+    return index;
+  }
+
```
dlabrecq: I'd like to see a new trackBy example created instead of altering the basic one. Want to ensure our basic use-case continues to work for existing applications

src/app/list/basic-list/example/list-compound-example.component.ts
```diff
@@ -155,11 +158,27 @@ export class ListCompoundExampleComponent implements OnInit {
...
+  }
+
+  setItems(): void {
+    this.items = cloneDeep(this.items);
+  }
+
+  getTrackBy(index: number, item: any): any {
+    return index;
+  }
+
```
dlabrecq: Please create one, new example for "trackBy". That will help ensure existing applications continue to work as is.

dlabrecq (changes requested):

seanforyou23:
Tested this against Syndesis UI and it works like a charm - nice enhancement

riccardo-forina:
Thanks for the review @dlabrecq! I have made some changes to the code, restored the examples and added a new one.

src/app/list/basic-list/example/list-polling-example.component.html
```diff
@@ -0,0 +1,159 @@
...
+  <div>
+    <tabset>
+      <tab heading="api">
+        <iframe class="demoframe" src="docs/classes/listcomponent.html"></iframe>
+      </tab>
+      <tab heading="html">
+        <include-content src="src/app/list/basic-list/example/list-basic-example.component.html"></include-content>
+      </tab>
+      <tab heading="typescript">
+        <include-content src="src/app/list/basic-list/example/list-basic-example.component.ts"></include-content>
```
dlabrecq: In order to show the correct example, these two links should use `list-polling-example.component.ts` instead of `list-basic-example.component.ts`.

src/app/list/basic-list/example/list-polling-example.component.ts
```diff
@@ -0,0 +1,194 @@
...
+      emptyStateConfig: this.emptyStateConfig,
+      multiSelect: false,
+      selectItems: false,
+      selectionMatchProp: 'name',
+      showCheckbox: true,
+      showRadioButton: false,
+      useExpandItems: true
+    } as ListConfig;
+
+    this.updateItemsInterval = setInterval(() => this.updateItems(), 2500);
```
dlabrecq: I'm seeing an error in my IDE, which states "Type 'Timer' is not assignable to type 'number'. However, using this seems to work:

`updateItemsInterval: number = <any>setInterval(() => {});`

See https://github.com/TypeStrong/atom-typescript/issues/1053

dlabrecq (changes requested):
Thank you for creating a new example. Just a couple nits left

src/app/list/basic-list/example/list-polling-example.component.html
riccardo-forina: ah! Good catch, thanks!

src/app/list/basic-list/example/list-polling-example.component.ts
riccardo-forina: I had it too but assumed it was something specific to my IDE (WebStorm). I'll update this to use `any` then.

dlabrecq (approved):
LGTM. Thanks for the contribution!

PR #439, opened 2018-08-06
add disabled type for filter dropdown

T0MASD:
This is good when a user wants to disable an option within the filter dropdown. It can be done by setting FilterField type value as FilterType.DISABLED


T0MASD:
It would be better to implement this using new property called disabled instead of type.

PR #440, opened 2018-08-06
feat(filter): add disabled property for filter dropdown.

T0MASD:
This is good when a user wants to disable an option within the filter dropdown. It can be done by setting FilterField disabled property to true


dlabrecq (approved):
Looks good.

PR #443, opened 2018-08-08
fix(tabs): ensure events are of appropriate type before setting tab

seanforyou23:
This PR adds an instanceof check for example components, to ensure events triggered from within a tabset are of the correct type before attempting to access heading property of the TabDirective event object.

This seems to be a bug in ngx-bootstrap as they use the `(select)` event binding which collides with the DOM event `select`.

More info here: https://github.com/angular/angular/issues/23603

ensure (select) event is instanceof TabDirective before setting heading
remove unused imports
add flag for project in tslint target


dlabrecq (approved):
LGTM

PR #444, opened 2018-08-08
feat(copy-updates): align copy component impl with latest designs

seanforyou23:
This PR updates the copy clipboard components to use the latest designs and interactions specified in [core](https://www.patternfly.org/pattern-library/forms-and-controls/copy-to-clipboard/).

One outstanding issue that I'm not sure how to resolve is that the tooltip text for the copy buttons updates to "Copied" for a few seconds after firing the copy action, however this value isn't updated in the DOM until you move your cursor away from the copy button and then back over the top of it again. I'm using property bound `[tooltip]` syntax which I thought would handle this but doesn't seem to. Any idea how to solve this?

make widths configurable
improve tests
add tooltip for copy buttons
specify tooltip container to fix tooltip flutter
use input element for copy value container
remove redundant triggers
clean up view templates
use obj for storing common values in copyBase
use readonly input instead of span so copy value is scrollable
use tooltipText input prop instead of tooltip to avoid namespace collision
specify body for container attribute to fix tooltip styling issues
avoid breaking changes by setting buttonAriaLabel to buttonLabel value when passed in
update tests to use new property name

### DIFF
--- a/src/app/copy/copy-base.ts
+++ b/src/app/copy/copy-base.ts
@@ -11,6 +11,11 @@ import { CopyService } from './copy-service/copy.service';
  * A config containing properties for copy components
  */
 export abstract class CopyBase {
+  protected _copyDefaults = {
+    copyBtnTooltipText: 'Copy to Clipboard',
+    tooltipCopiedText: 'Copied'
+  };
+
   /**
    * Copy button aria label (announced to screen readers)
    */
@@ -19,7 +24,7 @@ export abstract class CopyBase {
   /**
    * A tooltip that describes the value to be copied
    */
-  @Input('tooltip') tooltip: string;
+  @Input('tooltipText') tooltipText: string;
 
   /**
    * Placement for the tooltip
@@ -27,9 +32,29 @@ export abstract class CopyBase {
   @Input('tooltipPlacement') tooltipPlacement: string = 'top';
 
   /**
+   * A tooltip that describes what the copy button does
+   */
+  @Input('copyBtnTooltipText') copyBtnTooltipText: string = this._copyDefaults.copyBtnTooltipText;
+
+  /**
+   * A tooltip that informs user that copy action recently occurred
+   */
+  @Input('tooltipCopiedText') tooltipCopiedText: string = this._copyDefaults.tooltipCopiedText;
+
+  /**
+   * Placement for the copy button tooltip
+   */
+  @Input('copyBtnTooltipPlacement') copyBtnTooltipPlacement: string = 'top';
+
+  /**
    * The value to be copied to the clipboard
    */
   @Input('value') value: string;
+
+  /**
+   * A width to set on the copy container
+   */
+  @Input('width') width: string;
 
   /**
    * Event emitted when values are copied to the clipboard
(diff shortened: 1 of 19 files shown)


src/app/copy/copy-base.ts
```diff
@@ -11,6 +11,13 @@ import { CopyService } from './copy-service/copy.service';
  * A config containing properties for copy components
  */
 export abstract class CopyBase {
+  protected _copyDefaults = {
+    tooltipPlacement: 'top',
+    copyBtnTooltipPlacement: 'top',
+    copyBtnTooltipText: 'Copy to Clipboard',
+    copiedTooltipText: 'Copied'
+  };
```
dlabrecq: Please move the defaults to the corresponding inputs in this class. It looks like only "Copied" is used more than once?

src/app/copy/copy-base.ts
```diff
@@ -62,8 +84,10 @@ export abstract class CopyBase {
         value: this.value
       } as CopyEvent);
       this._recentlyCopied = true;
+      this.copyBtnTooltipText = this._copyDefaults.copiedTooltipText;
       setTimeout(() => {
         this._recentlyCopied = false;
+        this.copyBtnTooltipText = this._copyDefaults.copyBtnTooltipText;
```
dlabrecq: This is a little weird. The "Copied" tooltip appears to be set to the same default value twice; once on line 87, then on line 90 in the timeout? How is the tooltip restored to the original copyBtnTooltipText value provided by the developer?

`this.copyBtnTooltipText = this._copyDefaults.copiedTooltipText;`

Do we need to provide another tootltip property (e.g., tooltipCopiedText), so devs can localize their own "Copied" text?

src/app/copy/copy-base.ts
```diff
@@ -19,18 +26,33 @@ export abstract class CopyBase {
...
 
   /**
    * The value to be copied to the clipboard
    */
   @Input('value') value: string;
 
+  /**
+   * A width to set on the copy container
+   */
+  @Input('width') width: string;
```
dlabrecq, line 57: Is there a default width for the pfng-block-copy selector? Just wondering what happens when this property is not provided?

dlabrecq (changes requested):

src/app/copy/copy-base.ts
seanforyou23: Good catch! Yes I think adding a dedicated property for the copied msg is the way to go, adding that to my latest changes. Also, there wasn't a way to restore to the original copyBtnTooltipText, so adding that as well.

src/app/copy/copy-base.ts
seanforyou23: We're basically not setting a default width so the inline copy remains inline flow content out of the box.

Block copy is encased within a div which is block level by default so it will occupy 100% of available width in the parent element.

Because the width is a shared property between inline and block, not setting a default value here seemed to make most sense. I could set `width: 100%` on `.pfng-block-copy` if you think this will become a tripping hazard for devs.

dlabrecq: That is fine. I was more curious if setting width was a requirement

src/app/copy/copy-base.ts
seanforyou23: See latest updates, we're using 'Copy to Clipboard' and 'Copied' potentially twice so I left those in a defaults object, but if it still doesn't make sense - happy to move them out of here. Let me know what you think!

dlabrecq (dismissed):
LGTM

PR #445, opened 2018-08-09
Replaced NotificationModule by ToastNotificationModule

oveits:
If I follow the instructions in the README, I get an error 
```
"export 'NotificationModule' was not found in 'patternfly-ng'
```
since NotificationModule is not found. ToastNotificationModule is found, and fits the component added to the HTML template, i.e. `<pfng-toast-notification />` .


dlabrecq (approved):
Good catch. Thank you!

PR #449, opened 2018-08-22
fix(filter): Removed placeholder from within dropdown…

### LINKED ISSUE #418: Filter bug
This issue re: http://www.patternfly.org/patternfly-ng/#/filters was originally filed by @akarem on patternfly-design. See original issue here: https://github.com/patternfly/patternfly-design/issues/695

**Description of bug**

image

Here are the problems I see with this weekday filter.
First I choose Monday -> drop-down does not change it's state, but tag is appended 
image
Does it mean that I can add another weekday? It looks like yes, but If I add tuesday the tag is replaced.
image
Ok, I can't add another day. So this control acts like radio group / single-select dropdown / combobox.
What if I want to remove my selection? I click again and choose that initial option 
image
And nothing happens..
The only way to remove selected filter is to click on `x` in "Active filters" list.

**So the question: Why this default option even there? It does nothing.**

This component looks like classic combobox with placeholder but acts like button group where first button does not have any action.

It probably should look like this (when it does not remember it's state)
```
 +------------------------------------------------------+
 |  Week Day | V | Filter by week day...            | V |
 +------------------------------------------------------+
```
or like this (when it can remember)
```
 +------------------------------------------------------+
 |  Week Day | V | No filter                        | V |
 +------------------------------------------------------+
```
while not interactive.

And like this when drop-down is activated
```
 +------------------------------------------------------+
 |  Week Day | V |                                  | V |
 +------------------------------------------------------+
                 | No filter                            |
                 | Sunday                               |
                 | Monday                               |
                 | Tuesday                              |
                 | Wednesday                            |
                 | Thursday                             |
                 | Friday                               |
                 | Saturday                             |
                 +--------------------------------------+
```
If user chooses "No filter" option, then filter should be removed. 
If user clisks on something else, then crsp filter is added/replaced.

After user chosed anything, it should look like this (when it does not remember it's state)
```
 +------------------------------------------------------+
 |  Week Day | V | Filter by week day...            | V |
 +------------------------------------------------------+
```
or like this (when it can remember)
```
 +------------------------------------------------------+
 |  Week Day | V | No filter                        | V |
 +------------------------------------------------------+
```
```
 +------------------------------------------------------+
 |  Week Day | V | Monday                           | V |
 +------------------------------------------------------+
```
If you don't want to allow user to remove filters from dropdown, then you may remove "No filter option".

_____________

So, one more time..

If it's combobox, it's legal to add default value (placeholder) as first option, but it __must__ display currently selected option and if user selects default option it __must__ be applied

If it's drop-down button group or "menu", than it __must__ __not__ display menu placeholder, it __must__ __not__ contain any buttons that don't do anything.

dtaylor113:
… and show current value in dropdown

**Before:**
- Placeholder text within dropdown
- Selected/Applied filter value not reflected in dropdown
filter_before

**After:**
- Placeholder text removed from dropdown
- Selected/Applied filter value reflected in dropdown
- Dropdown remembers previously applied filter value
filter_after

CC @mcarrano,  @akaRem

Fixes #418

### DIFF
--- a/src/app/filter/filter-fields.component.html
+++ b/src/app/filter/filter-fields.component.html
@@ -29,15 +29,13 @@
       <div class="btn-group bootstrap-select form-control filter-select" dropdown>
         <button type="button" class="btn btn-default dropdown-toggle" dropdownToggle
                 [disabled]="config.disabled === true">
-          <span class="filter-option pull-left">{{currentValue || currentField?.placeholder}}</span>
+          <span class="filter-option pull-left"
+                [ngClass]="{'placeholder': !currentValue}">
+            {{currentValue || currentField?.placeholder}}
+          </span>
           <span aria-hidden="true" class="caret"></span>
         </button>
         <ul class="dropdown-menu" role="menu" *dropdownMenu>
-          <li role="menuitem" *ngIf="currentField?.placeholder">
-            <a class="dropdown-item" href="javascript:void(0);" tabindex="-1" (click)="selectQuery()">
-              {{currentField?.placeholder}}
-            </a>
-          </li>
           <li role="menuitem" *ngFor="let query of currentField?.queries"
               [ngClass]="{'selected': query?.value === currentValue, 'divider dropdown-divider': query?.separator}">
             <a class="dropdown-item" href="javascript:void(0);" tabindex="-1"
@@ -70,11 +68,6 @@
           <span (click)="queryInput.focus()" class="caret"></span>
         </div>
         <ul class="dropdown-menu" role="menu" *dropdownMenu>
-          <li role="menuitem" *ngIf="currentField.placeholder">
-            <a class="dropdown-item" href="javascript:void(0);" tabindex="-1" (click)="selectQuery()">
-              {{currentField?.placeholder}}
-            </a>
-          </li>
           <li role="menuitem" *ngFor="let query of currentField?.queries"
               [ngClass]="{'selected': query.value === currentValue,
                           'divider dropdown-divider': query?.separator,
--- a/src/app/filter/filter-fields.component.less
+++ b/src/app/filter/filter-fields.component.less
@@ -33,9 +33,11 @@
   .btn-default {
     background-color: @color-pf-white;
     background-image: none;
-    color: @color-pf-black-500;
-    font-style: italic;
-    font-weight: 400;
+    .placeholder {
+      color: @color-pf-black-500;
+      font-style: italic;
+      font-weight: 400;
+    }
   }
   .avatar {
     height: 20px;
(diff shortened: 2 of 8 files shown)


mcarrano:
@dtaylor113 This looks promising. But I cannot seem to get to the preview link. Can you check it?

dlabrecq:
@mcarrano Try the link again. It worked for me, but takes a bit to load

mcarrano:
This looks great, @dtaylor113 . My only question is whether it can be made so that the text is only italicized when it is the placeholder. At all other times, the selection should not be italicized. If this is not possible, then I think I would remove the italicized style altogether. 

As for the multi-select variant we talked about yesterday, here is a quick mockup to show what this could look like. Let me know what you think.

multi-select filter

dtaylor113:
I was thinking the same thing @mcarrano :-)  Here's the updated screenshots:
image
image

dlabrecq:
@mcarrano I may have encountered an issue with the pattern.

1. Choose a birth month option -- the selection is shown in the menu
2. Now change the filter field from `Birth Month` to `Name`
3. Change the filter field back to `Birth Month`

Although the filter has been applied, the filter selection is no longer reflected in the menu. Wondering if the menu selection is expected to reflect the applied filter in this scenario?

new gif

dtaylor113:
> Although the filter has been applied, the filter selection is no longer reflected in the menu

Hi @dlabrecq, I raised this in the middle of https://github.com/patternfly/patternfly/issues/1116.  I'm still not sure how complicated it will be to have the dropdown remember it's state -especially if we eventually support multiple select values. It also depends on whether we also support enforcing single select or allow dropdown filters to be accumulative.

I'll spend a little more time looking into this issue.  I think for now we are always enforcing single select/applied filter for dropdowns.  Meaning a dropdown always replaces an existing filter tag.  In this scenario it makes sense to reset the dropdown to the applied filter value.

dtaylor113:
@mcarrano, @dlabrecq Dropdowns now remember previously applied filter values.  I've updated this PRs Description and screenshots to show this.  -thanks

dlabrecq:
That's fine. Mostly curious if it's going to be an issue or not

mcarrano:
This is looking good @dtaylor113 Do you want to try to incorporate the multi-select here also or would you rather consider that in a separate PR?

dtaylor113:
> This is looking good @dtaylor113 Do you want to try to incorporate the multi-select here also or would you rather consider that in a separate PR?

Separate PR please!  I need to apply these changes to patternfly-react next, then I'd work on multi-select.  -thanks

@mcarrano, @dlabrecq please approve this PR when/if ready -thanks

dlabrecq (approved):
LGTM

mcarrano (approved):
Looks great @dtaylor113 . I will open a separate issue for the multi-select version of this control.

dtaylor113:
> I will open a separate issue for the multi-select version of this control.

Thanks @mcarrano, please be sure to indicate that multi-select should use 'Grouped Filter Chips'.

Issue #451, opened 2018-12-13
Typo in NotificationGroup


LarsLuz:
Hello!

I found a typo in the file "notification-group.ts" - the class name is written as "NotificaitonGroup".

Thank you for your time!

joshuawilson:
It is not consistently misspelled either. :(

joshuawilson:
@LarsLuz you should submit a PR to fix it.

LarsLuz:
I hope I didn't overlook any of the typos.

dlabrecq:
I believe this was fixed via https://github.com/patternfly/patternfly-ng/pull/452

PR #452, opened 2018-12-14
Changend class name of "NotificaitonGroup" to "NotificationGroup" (#451)

LarsLuz:


src/app/notification/index.ts
```diff
@@ -1,6 +1,6 @@
 export { Notification } from './notification';
 export { NotificationEvent } from './notification-event';
-export { NotificaitonGroup } from './notification-group';
```
dlabrecq, line 3: Although trivial, this may be a breaking change? In addition to these modifications, I'm thinking we should add the export below to avoid breaking existing apps.

export { NotificationGroup as NotificaitonGroup } from './notification-group'; // deprecated

dlabrecq (changes requested):
Not sure why the CI build failed, but have restarted it

src/app/notification/index.ts
joshuawilson: I was wondering if that might break things. I think this is a good way to handle that.

LarsLuz: Sorry I've misread the first post.

dlabrecq:
@LarsLuz It appears that a test is failing.

https://travis-ci.org/patternfly/patternfly-ng/builds/469072033#L1270

LarsLuz:
It seems like a test of the donut chart is failing. I do not really understand why, it seems unrelated to the name change of NotifiicationGroup.

The offending test:
https://travis-ci.org/patternfly/patternfly-ng/builds/469072033#L1129

dlabrecq:
The tests fail locally, but commits 8476209 and ec28940 work.

It appears that the last change is causing problems. We may not be able to define NotificationGroup as two different things in the same index.ts file?

`export { NotificationGroup as NotificaitonGroup } from './notification-group'`

LarsLuz:
We could also create a stub class with the old name and let it inherit NotificationGroup?
What do you think?

dlabrecq:
Sounds good idea to me. I was thinking about using a stub class -- and marking it deprecated.

dlabrecq (approved):

Issue #475, opened 2024-04-25
Unable to install packages based on package-lock file 


soGit:
```bash
npm ERR! 404 Not Found - GET https://registry.npmjs.org/es-abstract/-/es-abstract-1.14.0.tgz
npm ERR! 404
npm ERR! 404 es-abstract@1.14.0 is not in the npm registry.
```

Seems like es-abstract unpublished v1.14.0

dlabrecq:
Unfortunately, this project has not been supported for some time. The last commit was 5 yrs ago and it's unlikely a new version will be published.

That said, I've updated the README to indicate this project is no longer active. Will discuss archiving the repo with team owners

PR #476, opened 2024-04-25
fix(dependencies): fix package lock dependencies

soGit:
Recreate package-lock due to issue #475

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -68,7 +68,7 @@
         "arr-diff": {
           "version": "2.0.0",
           "resolved": "https://registry.npmjs.org/arr-diff/-/arr-diff-2.0.0.tgz",
-          "integrity": "sha1-jzuCf5Vai9ZpaX5KQlasPOrjVs8=",
+          "integrity": "sha512-dtXTVMkh6VkEEA7OhXnN1Ecb8aAGFdZ1LFxtOCoqj4qkyOJMt7+qs6Ahdy6p/NQCPYsRSXXivhSB/J5E9jmYKA==",
           "dev": true,
           "requires": {
             "arr-flatten": "^1.0.1"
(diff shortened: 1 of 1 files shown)


dlabrecq:
Curious how did you create the package-lock.json file? If I remove this file, I cannot reproduce it using the existing package.json. Did you for get to check-in your package.json?

Screenshot 2024-04-25 at 2 15 35 PM

soGit:
> Curious how did you create the package-lock.json file? If I remove this file, I cannot reproduce it using the existing package.json. Did you for get to check-in your package.json?
> 
> Screenshot 2024-04-25 at 2 15 35 PM


Hello, let me clarify. If you open package-lock.json on master branch, you will find that es-abstract@1.14.0 version was resolved 

package-lock

but this version does not exist anymore.

npm-es-abstract
You can try to run `npm i` without deleting package-lock.json. 404 error should appear
404-on-install



So I have recreated package-lock.json and now es-abstract@1.23.3 version resolved.

package-lock

dlabrecq:
It's probably not a good idea to edit package-lock.json directly. We should be able to delete package-lock.json and regenerate it from a fresh `npm install`. Otherwise, the lock file won't be reproducible.

In an attempt to regenerate the package-lock file myself, I added the following to package.json. Then, I could delete package-lock.json and run `npm install`, which resolved to es-abstract@1.23.3.

```
  "overrides": {
    "css-loader": "1.0.0",
    "stylelint": "9.4.0",
    "typedoc": "0.11.1",
    "typescript": "2.9.2"
  }
```

Unfortunately, this project has not been supported for some time. The last commit was 5 yrs ago.

We appreciate fixes, but the project no longer builds locally and the UI won't deploy -- Travis isn't even building this PR. It's unlikely we will publish a new package version at this time. Thank you.

soGit:
> It's probably not a good idea to edit package-lock.json directly. We should be able to delete package-lock.json and regenerate it from a fresh `npm install`. Otherwise, the lock file won't be reproducible.
> 
> In an attempt to regenerate the package-lock file myself, I added the following to package.json. Then, I could delete package-lock.json and run `npm install`, which resolved to es-abstract@1.23.3.
> 
> ```
>   "overrides": {
>     "css-loader": "1.0.0",
>     "stylelint": "9.4.0",
>     "typedoc": "0.11.1",
>     "typescript": "2.9.2"
>   }
> ```
> 
> Unfortunately, this project has not been supported for some time. The last commit was 5 yrs ago.
> 
> We appreciate fixes, but the project no longer builds locally and the UI won't deploy -- Travis isn't even building this PR. It's unlikely we will publish a new package version at this time. Thank you.
> 

I did not manually change the lock file. I deleted it and ran npm install to recreate lock file.

You don't really need to change the package.json, the problem with incompatible package versions that you are facing is a different one. You can use the old node (for me version 11 works well for this project).

By the way, I successfully ran a project locally. With Python 2.7 and node 11

What do you think is better? Make a public archive (close the repo) or try to update CI/CD?

dlabrecq:
I now realize the package-lock issue is related to the version of node used to create the lock file. When using v18.x, I'm not able to recreate the lock file without modifying package.json. I can regenerate the lock file using node v11.x; however, I still could not build or deploy the UI.

That said, I've updated the README to indicate this project is no longer active. Will discuss archiving the repo with team owners

soGit:
@dlabrecq sorry, I've checked that wrong version was pushed.
I've created new MR. https://github.com/patternfly/patternfly-ng/pull/477
Could you please check one more time with node v11 and python v2.7?

dlabrecq:
I did test with node 11.x (above), but encountered aot build failures while the UI generates d3 / minimatch typescript errors. I also re-tested with node v8.x, which is the minimum requirement in package.json. In addition, I retested MR #477.

That said, I don't believe the errors have much to with the lock file. I encounter the same errors after rebuilding the lock file myself. However, it would require significant effort to revive this project in order to test properly.

PR #477, opened 2024-04-29
fix(dependencies): fix package lock dependencies

soGit:
Recreate package-lock due to issue https://github.com/patternfly/patternfly-ng/issues/475

IMPORTANT: Use node v 11 & python v 2.7


dlabrecq:
Tested with node 11.x, but encountered aot build failures while the UI generates d3 / minimatch typescript errors.

I don't believe the build errors have much to with the lock file. I encounter the same errors after rebuilding the lock file myself. However, it would require significant effort to revive this project in order to test properly.

Thank you for the contribution, but this project is no longer supported. Without being able to test the changes against the UI, I will have to close this PR.

